PR #7, opened 2020-04-29
Ra dspdc 982 refactor hles

raaidbroad:
Export raw values instead of labels, export in "eav" mode instead of "flat" mode, establish a global concurrency limit like in ENCODE.

### DIFF
--- a/hle-survey/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilder.scala
+++ b/hle-survey/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilder.scala
@@ -86,7 +86,7 @@ class HLESurveyExtractionPipelineBuilder(
       .parallelize(Iterable(()))
       .transform("Get study IDs") {
         _.applyKvTransform(ParDo.of(idLookupFn)).flatMap { kv =>
-          kv.getValue.fold(throw _, _.arr.map(_.read[String]("study_id")))
+          kv.getValue.fold(throw _, _.arr.map(_.read[String]("value")))
         }
       }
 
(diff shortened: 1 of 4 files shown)


hle-survey/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilder.scala
```diff
@@ -61,6 +63,7 @@ class HLESurveyExtractionPipelineBuilder(
 
   override def buildPipeline(ctx: ScioContext, args: Args): Unit = {
     import org.broadinstitute.monster.common.msg.MsgOps
+    import HLESurveyExtractionPipelineBuilder._
 
     val idLookupFn = new ScalaAsyncLookupDoFn[Unit, Msg, RedCapClient]() {
```
raaidbroad, line 68: hold up; should this `ScalaAsyncLookupDoFn` also have the `MaxConcurrentRequests` parameter passed to it?

danxmoran: Oh, urgh.

Ideally we'd have a single lookup fn instance w/ the parameter set, so the two types of lookup share the same resources. That's not a big deal here, since this call is made exactly 1 time at workflow start, but it'll get more important once we start downloading data dictionaries too.

The way I would probably model it would be to introduce a new ADT:
```scala
sealed trait RedcapRequest
case class GetRecords(fields: List[String], forms: List[String], start: Option[OffsetDateTime], end: Option[OffsetDateTime], filters: Map[String, String]) extends RedcapRequest
# Future: case class GetDataDictionary(instrument: String)
```
Then the solo lookup FN takes an instance of `RedcapRequest`, and calls the right thing on the `client` it wraps depending on the arg it pulls.

I don't feel strongly about whether that change is made in this PR or in the follow-up to pull data dictionaries. If you want to punt on it, I think it's fine to merge as-is.

raaidbroad: Awesome, I'll throw that change in to the data dictionary ticket I'm doing. Thanks!

PR #10, opened 2020-05-04
DSPDC-1040 Fix query for data dictionaries.

danxmoran:
Took the chance to nitpick some names / constants

### DIFF
--- a/hle-survey/extraction/src/it/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilderIntegrationSpec.scala
+++ b/hle-survey/extraction/src/it/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilderIntegrationSpec.scala
@@ -5,6 +5,7 @@ import java.time.{LocalDate, LocalTime, OffsetDateTime, ZoneOffset}
 import better.files.File
 import com.bettercloud.vault.{SslConfig, Vault, VaultConfig}
 import org.broadinstitute.monster.common.PipelineBuilderSpec
+import org.broadinstitute.monster.common.msg.JsonParser
 
 class HLESurveyExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
   val outputDir = File.newTemporaryDirectory()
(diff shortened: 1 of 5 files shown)


hle-survey/extraction/src/it/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -53,6 +54,16 @@ class HLESurveyExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderS
   }
 
   it should "successfully download data dictionaries from RedCap" in {
-    readMsgs(outputDir, "data_dictionaries/*.json") shouldNot be(empty)
+    readMsgs(outputDir, "data-dictionaries/*.json") shouldNot be(empty)
+  }
+
+  it should "not download data dictionaries multiple times" in {
```
raaidbroad, line 60: lolololol

hle-survey/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipelineBuilderSpec.scala
```diff
@@ -88,6 +88,6 @@ class HLESurveyExtractionPipelineBuilderSpec extends PipelineBuilderSpec[Args] {
   }
 
   it should "write downloaded data dictionaries to disk" in {
-    readMsgs(outputDir, "data_dictionaries/*.json") shouldBe expectedDataDictionary.toSet
+    readMsgs(outputDir, "data-dictionaries/*.json") shouldBe expectedDataDictionary.toSet
```
snwessel, line 91: We do use the underscore in the output directories for Clinvar and Encode... but otherwise this looks good to me. I don't really have a preference 🤷

danxmoran: Ah, that's fair... In those cases we're mirroring the target table names (not the case for extraction), but no point in being inconsistent. I'll revert.

PR #12, opened 2020-05-11
Sw dspdc 850 transform owner

snwessel:
Transforming the owner object. Also:
- Adding a helper method to read boolean values
- Adding a column to the owner table that had been left out by mistake. 

The mapping doc is here: https://docs.google.com/spreadsheets/d/1QqWzkVtsUUjOvOqCJEA8R6YcfcAq87MzYvMLnMQBa58/edit?usp=sharing


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -6,19 +6,32 @@ import com.spotify.scio.values.SCollection
...
   case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
+    val logger = LoggerFactory.getLogger(getClass)
 
     def getRequired(field: String): String = fields(field).head
 
     def getOptional(field: String): Option[String] = fields.get(field).flatMap(_.headOption)
 
-    def getArray(field: String): Array[String] = fields.get(field).getOrElse(Array.empty)
+    def getBooleanOption(field: String): Option[Boolean] = {
+      val rawFieldValue = this.getOptional(field)
```
danxmoran, line 24: `this` not needed

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -6,19 +6,32 @@ import com.spotify.scio.values.SCollection
...
+      if (rawFieldValue.isEmpty) None
+      else if (rawFieldValue.head == "1") Some(true)
+      else if (rawFieldValue.head == "0") Some(false)
```
danxmoran, line 27: Since it's going back to an `Option` anyways, more conventional would be to `rawFieldValue.map(_ == "1")`

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -6,19 +6,32 @@ import com.spotify.scio.values.SCollection
...
+      else {
+        logger.warn(s"Invalid yes/no value in field $field on record $id: '$rawFieldValue'")
```
danxmoran, line 29: I don't think it's worth logging here, there are some cases where "Unknown" is a valid response. I think they use "99" or something crazy to represent that

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -6,19 +6,32 @@ import com.spotify.scio.values.SCollection
...
 
   case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
+    val logger = LoggerFactory.getLogger(getClass)
 
     def getRequired(field: String): String = fields(field).head
 
     def getOptional(field: String): Option[String] = fields.get(field).flatMap(_.headOption)
 
-    def getArray(field: String): Array[String] = fields.get(field).getOrElse(Array.empty)
+    def getBooleanOption(field: String): Option[Boolean] = {
```
danxmoran, line 23: Could this short-circuit and just return `Boolean`? Instead of `map` below, you'd use `contains`. I think that will cover the most common pattern we need in the data

danxmoran: (The most common pattern being a check to see if a Y/N question is "Yes" before extracting a bunch of other fields)

snwessel: Yeah, that makes sense. I had made it return an option for fields like `od_hispanic` that we have in the schema as non-required booleans. For that field,  it feels a little weirder to default to false. What do you think?

snwessel: @danxmoran  I simplified things with your suggestions, but left a "getBooleanOption" method for now to be used for fields like `od_hispanic`, where we don't want to default to false if the field was not filled out. Let me know what you think. I am happy to change it if you disagree.

danxmoran: I think returning the `Option` optimizes for storing the result as a column, and returning a raw `Boolean` optimizes for using the Y/N to drive other transformations. Here there's 1 of each case, but I think the Y/N driving will be much more common overall.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -6,19 +6,25 @@ import com.spotify.scio.values.SCollection
...
 import org.broadinstitute.monster.dogaging.jadeschema.table.{HlesDog, HlesOwner}
+import org.slf4j.LoggerFactory
 import upack.Msg
 
 object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
 
   implicit val msgCoder: Coder[Msg] = Coder.beam(new UpackMsgCoder)
 
   case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
+    val logger = LoggerFactory.getLogger(getClass)
```
danxmoran, line 17: Is this used anywhere?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
danxmoran: Having both options works for me!

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
snwessel: Ope, not anymore! Good catch. Just removed it.

PR #15, opened 2020-05-22
Sw dspdc 1081 update and unit test owner transformation

snwessel:
Updated the owner transformation to reflect the "unfurling" changes to the schema and added a couple of unit tests. Let me know if there are other tests you think I should include. 

Here is the mapping doc for reference: https://docs.google.com/spreadsheets/d/1QqWzkVtsUUjOvOqCJEA8R6YcfcAq87MzYvMLnMQBa58/edit?usp=sharing


schema/src/main/jade-tables/hles_owner.table.json
```diff
@@ -33,35 +60,35 @@
...
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "oc_primary_residence_zip",
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "oc_primary_residence_ownership",
```
snwessel, line 94: Ope, I just realized that this should be an integer... I'll fix that :)

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -43,6 +43,13 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
...
   def getOptionalBoolean(field: String): Option[Boolean] = getOptional(field).map(_ == "1")
 
+  /** Get the singleton value for an attribute in this record, parsed as a long. */
+  def getOptionalNumber(field: String): Option[Long] = getOptional(field).map(_.toLong)
+
   /** Get every value for an attribute in this record. */
   def getArray(field: String): Array[String] = fields.getOrElse(field, Array.empty)
+
+  /** If the field exists, check whether it contains the value, otherwise return None. */
+  def containsValue(field: String, value: String): Option[Boolean] =
```
danxmoran, line 53: I'm not sure about the utility of this helper. It runs the `getOrElse` per-call, and forces `Option[Boolean]` (which will be the most common case, but if we ever need `Boolean` we'll need to redo it).

I'd rather have the mapping code use `getArray` to pull the values once, then call `contains` for each option.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/OwnerTransformationsSpec.scala
```diff
@@ -0,0 +1,81 @@
...
+    "oc_address1_division" -> Array("3"),
+    "oc_address1_zip" -> Array("01111"),
+    "oc_address1_own" -> Array("1"),
+    "oc_address1_own_other" -> Array("some text"),
+    "oc_address2_yn" -> Array("1"),
+    "oc_address2_state" -> Array("MA"),
+    "oc_address2_zip" -> Array("02222"),
+    "oc_address2_own" -> Array("2"),
+    "oc_address2_own_other" -> Array("some text")
+  )
```
danxmoran, line 31: Can we also include a test case where all of the optional things are missing?

schema/src/main/jade-tables/hles_owner.table.json
```diff
@@ -33,39 +60,39 @@
...
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "oc_primary_residence_ownership_other",
```
danxmoran, line 98:
```suggestion
      "name": "oc_primary_residence_ownership_other_description",
```
New convention for the free-text details of "other" answers

schema/src/main/jade-tables/hles_owner.table.json
```diff
@@ -77,11 +104,11 @@
...
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "oc_secondary_residence_ownership",
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "oc_secondary_residence_ownership_other",
```
danxmoran, line 114:
```suggestion
      "name": "oc_secondary_residence_ownership_other_description",
```

schema/src/main/jade-tables/hles_owner.table.json
```diff
@@ -8,23 +8,50 @@
...
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "od_max_education",
-      "datatype": "string"
+      "datatype": "integer"
     },
     {
       "name": "od_max_education_other",
```
danxmoran, line 18:
```suggestion
      "name": "od_max_education_other_description",
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
snwessel: Okay, sounds good. I agree that it's a bit redundant. We do want it to return `None` when the field is not defined though, right? It looks like doing `getArray(...).contains(...)` would return false when the field is not defined in the RawRecord.

danxmoran: To clarify, I think this method should go away entirely (so "it" won't return anything anymore).

Whether or not the end value is wrapped in an `Option` will depend on the use-site. The simplest thing will be to wrap the values in `Some` when filling in the `case class`

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
```diff
@@ -7,32 +7,41 @@ object OwnerTransformations {
...
       ownerId = rawRecord.id,
-      odAgeRangeYears = rawRecord.getOptional("od_age"),
-      odMaxEducation = rawRecord.getOptional("od_education"),
-      odMaxEducationOther = rawRecord.getOptional("od_education_other"),
-      odRace = rawRecord.getArray("od_race"),
-      odRaceOther = rawRecord.getOptional("od_race_other"),
+      odAgeRangeYears = rawRecord.getOptionalNumber("od_age"),
+      odMaxEducation = rawRecord.getOptionalNumber("od_education"),
+      odMaxEducationOtherDescription = rawRecord.getOptional("od_education_other"),
+      odRaceWhite = raceValues.map(_.contains("1")),
```
snwessel, line 18: @danxmoran I think you had suggested doing something like `Some(rawRecord.getArray("od_race").contains("1"))` to get these values but I'm concerned that doing so would set the value to `Some(false)` when it should be undefined, so I implemented it a bit differently. Let me know if you disagree or think there is a better way of doing it.

danxmoran: I like it!

PR #16, opened 2020-05-22
Sw dspdc 1063 transform st columns

snwessel:
Adding the study status columns to the dog aging transformation (just 5 columns). I was using this as an excuse to also:
- get rid of the massive HlesDog constructor, and use the new init method instead
- set up a test class for the dog transformations

@danxmoran (after the weekend) let me know if there is something else you had in mind for how to test these transformations. I would be happy to change, especially if something else would be easier in the long run.

### DIFF
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
@@ -31,7 +31,8 @@ object DogTransformations {
   }
 
   /**
-    *
+    * Parse all study-status-related fields out of a raw RedCap record,
+    * injecting them into a partially-modeled dog record.
     */
   def mapStudyStatus(rawRecord: RawRecord, dog: HlesDog): HlesDog =
     dog.copy(
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
@@ -9,73 +9,10 @@ object DogTransformations {
 
   /** Parse all dog-related fields out of a raw RedCap record. */
   def mapDog(rawRecord: RawRecord): HlesDog = {
-    val dogBase = HlesDog(
+    val dogBase = HlesDog.init(
       dogId = rawRecord.id,
       // FIXME: Once DAP figures out a name for a dedicated owner ID, use that.
-      ownerId = rawRecord.id,
-      // withBreed
-      ddBreedPure = None,
-      ddBreedPureNonAkc = None,
-      ddBreedMixedPrimary = None,
-      ddBreedMixedSecondary = None,
-      // withAge
-      ddAgeYears = None,
-      ddAgeBasis = None,
-      ddAgeExactSource = Array.empty,
-      ddAgeExactSourceOther = None,
-      ddAgeEstimateSource = Array.empty,
-      ddAgeEstimateSourceOther = None,
-      ddBirthYear = None,
-      ddBirthMonth = None,
-      // withSex
-      ddSex = None,
-      // withSpayNeuter
-      ddSpayedOrNeutered = None,
-      ddSpayOrNeuterAge = None,
-      ddSpayMethod = None,
-      ddEstrousCycleExperiencedBeforeSpayed = None,
-      ddEstrousCycleCount = None,
-      ddHasBeenPregnant = None,
-      ddHasSiredLitters = None,
-      ddLitterCount = None,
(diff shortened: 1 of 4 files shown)


schema/src/main/jade-tables/hles_dog.table.json
```diff
@@ -17,6 +17,26 @@
         }
       ]
     },
+    {
+      "name": "st_vip_or_staff",
+      "datatype": "string"
```
snwessel, line 22: I think I might have been getting ahead of the updates to the mapping doc... I'm not sure if this column should actually be an integer.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
```diff
@@ -9,76 +9,14 @@ object DogTransformations {
 
   /** Parse all dog-related fields out of a raw RedCap record. */
   def mapDog(rawRecord: RawRecord): HlesDog = {
-    val dogBase = HlesDog(
+    val dogBase = HlesDog.init(
       dogId = rawRecord.id,
       // FIXME: Once DAP figures out a name for a dedicated owner ID, use that.
-      ownerId = rawRecord.id,
```
danxmoran, line 15: Glorious

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -43,6 +46,10 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
   /** Get the singleton value for an attribute in this record if one exists, parsed as a boolean. */
   def getOptionalBoolean(field: String): Option[Boolean] = getOptional(field).map(_ == "1")
 
+  /** Get the singleton value for an attribute in this record if one exists, parsed as a date. */
+  def getOptionalDate(field: String): Option[LocalDate] =
+    getOptional(field).map(LocalDate.parse(_, DateTimeFormatter.ofPattern("MM-dd-yyyy")))
```
danxmoran, line 51: Need to check some example data to be sure this date format will actually work.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/DogTransformationsSpec.scala
```diff
@@ -0,0 +1,29 @@
...
+import java.time.LocalDate
+
+class DogTransformationsSpec extends AnyFlatSpec with Matchers {
+  behavior of "DogTransformations"
+
+  // TODO add tests for dog demographics transformation
+
+  it should "correctly map study status fields" in {
+    val exampleDogFields = Map[String, Array[String]](
+      "dd_us_born" -> Array("1"), // a required field
```
danxmoran, line 14: Why is this required here?

schema/src/main/jade-tables/hles_dog.table.json
danxmoran: Yes, you got ahead of me! This one should be an int, the rest should be ok as-is.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/DogTransformationsSpec.scala
snwessel: It looks like one of the transformation methods uses getRequired for that field without being wrapped in a conditional. I had assumed it was intentional but maybe not? (the getRequired call is [here](https://github.com/DataBiosphere/dog-aging-ingest/blob/efc2a9f8f7acb2a3b34901412f1b23911478b831/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala#L257)).

danxmoran: Ah! Would change this test to call `mapStudyStatus` instead, so we can remove this.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
danxmoran: Weirdness... I can't actually find any examples of date files in extraction outputs. It's been a few weeks since I ran it so things might have updated in RedCap, @snwessel have you seen any examples IRL?

snwessel:
@danxmoran no I haven't been able to find an actual example of a date yet either (although I only have a subset of the data). I was basing that format off of the redcap field description, quoted here:
```
"select_choices_or_calculations":"if(((datediff(\"02-29-2000\",[dd_form_date],\"d\",\"mdy\",true) - ...
```

PR #19, opened 2020-05-29
DSPDC-1080 Refactor demographics mappings, and add tests.

danxmoran:
Still a huge diff, but half of it is tests!


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
```diff
@@ -40,137 +38,155 @@ object DogTransformations {
       stBatchLabel = rawRecord.getOptional("st_batch_label"),
```
danxmoran, line 38: Based on the size of this file, I think we'll want to start splitting up the mapping functions by source instrument. I could get started here to make the precedent by moving `mapStudyStatus` to its own file (and moving its test to a corresponding separate spec). What do you all think?

schema/src/main/jade-tables/hles_dog.table.json
```diff
@@ -243,22 +334,175 @@
...
+    {
+      "name": "dd_alternate_recent_residence_13_state",
+      "datatype": "string"
+    },
+    {
+      "name": "dd_alternate_recent_residence_13_zip",
+      "datatype": "string"
+    },
+    {
+      "name": "dd_alternate_recent_residence_13_weeks",
```
quazi-broad, line 504: Do we know we won't see more than 13 alternate addresses? Or is that just the highest you've seen so in the data so far?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
quazi-broad: I think that would be a good way to start the process of splitting up the mapping logic. So all mapping/logic for "st_" fields would follow?

schema/src/main/jade-tables/hles_dog.table.json
danxmoran: The RedCap survey itself encodes 13 as the upper bound (which is why we have to read differently-named raw fields for each). If it was truly unbounded, we'd have no choice but to use an array here.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
snwessel: I agree, that would be a good way of splitting it up 👍 this file is big enough already, and it wouldn't feel good to keep adding to it.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/DogTransformationsSpec.scala
```diff
@@ -5,28 +5,601 @@ import org.scalatest.flatspec.AnyFlatSpec
...
+  it should "map study status fields" in {
     val exampleDogFields = Map[String, Array[String]](
       "st_vip_or_staff" -> Array("2"),
       "st_batch_label" -> Array("this is my label"),
-      "st_invite_to_portal" -> Array("05-22-2020"),
-      "st_portal_account_date" -> Array("01-01-2000"),
-      "st_dap_pack_date" -> Array("12-31-1999")
+      "st_invite_to_portal" -> Array("2020-05-22"),
+      "st_portal_account_date" -> Array("2000-01-01"),
+      "st_dap_pack_date" -> Array("2020-01-15 10:21")
```
snwessel, line 24: Oh, so they are the standard format! I take it you were able to find an example of one of the dates then?

danxmoran: Yes, there was a bug on the RedCap side which flagged all dates as PII. They fixed it after our last call with them.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
danxmoran: Done

PR #27, opened 2020-06-11
Sw dspdc 1061 transform physical activity columns

snwessel:
Transforming the physical activity (pa) columns. 

Here is the[ mapping doc](https://docs.google.com/spreadsheets/d/1QqWzkVtsUUjOvOqCJEA8R6YcfcAq87MzYvMLnMQBa58/edit?usp=sharing).


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/PhysicalActivityTransformations.scala
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+    * Parse weather and outdoor surface fields out of a raw RedCap record,
+    * injecting the data into a partially-modeled dog record.
+    */
+  def mapWeather(rawRecord: RawRecord, dog: HlesDogPhysicalActivity): HlesDogPhysicalActivity = {
+    // moderate weather
+    val moderateMonths = rawRecord.getOptionalNumber("pa_warm_months")
+    val moderateDaily =
+      if (moderateMonths.isDefined && moderateMonths.head > 0)
```
danxmoran, line 51:
```suggestion
      if (moderateMonths.exists(_ > 0))
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/PhysicalActivityTransformations.scala
```diff
@@ -0,0 +1,274 @@
...
+        rawRecord.getOptionalNumber("pa_warm_outdoors")
+      else None
+    val hasModerateWeather = moderateDaily.isDefined && moderateDaily.head != 5L
```
danxmoran, line 54:
```suggestion
    val hasModerateWeather = moderateDaily.exists(_ != 5L)
```

danxmoran: And same for other uses of this pattern below

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/PhysicalActivityTransformations.scala
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+    * Calculate the percentage of time a dog walks at a given pace.
+    * @param selectedPaceTypes The set of all paces a dog walks at.
+    * @param paceType The pace for which we want to calculate a percent value
+    * @param pacePercent The proportion of time a dog walks at the given pace (only defined when the pace was one of multiple values selected).
+    * @return A decimal value representing the percentage of time a dog walks at the given pace.
+    */
+  def transformPace(
+    selectedPaceTypes: Option[Array[String]],
```
danxmoran, line 156: No need for this to be an `Option` based on the body. It can just be an `Array`, and the caller has to deal with it.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/PhysicalActivityTransformations.scala
```diff
@@ -0,0 +1,274 @@
...
+    paceType: String,
+    pacePercent: Option[Long]
+  ): Option[Double] =
+    if (selectedPaceTypes.isEmpty) { None }
```
danxmoran, line 160: If you do keep the arg as an `Option`, this `if`/`else` could be more idiomatic as a `flatMap`
```scala
selectedPaceTypes.flatMap { selectedPaces =>
  if (!selectedPaces.contains(paceType)) {
    // foo
  } else if (/* bar */) {
    // qux
  } else {
    // woo
  }
}
```

PR #28, opened 2020-06-11
QH - DSPDC-1118 - owner_primary_residence_census_division transformations

quazi-broad:
Updated hles_owner and hles_dog tables so that oc_primary_residence_census_division is an integer. Added transformation logic to parse the Division from the full provided string. Updated the unit test outputs to match this new logic.

### DIFF
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
@@ -16,6 +16,28 @@ object OwnerTransformations {
     val secondaryAddressOwnership = secondaryAddress.flatMap {
       if (_) rawRecord.getOptionalNumber("oc_address2_own") else None
     }
+    
+    // Parsing the oc_primary_residence_census_division
+    // format is: "Division <N>: <some-description>"
+    val censusDivisionPattern: Regex = "(Division \\d+:)"
+    
+    /**
+      * Extract the Division value from the full text string.
+      *
+      * @param divisionString the text string to be parsed
+      * @return an integer of the division number
+    **/
+    def getCensusDivision(divisionString: String): (integer) = {
+      val matches = censusDivisionPattern
+        .findFirstMatchIn(divisionString)
+        .getOrElse(
+          throw new Exception(
+            s"ownerTransformations: error while parsing census division id from $divisionString"
+          )
+        )
+      val censusDivisionId = matches.group(1)
+      censusDivisionId
+    }
 
     HlesOwner(
       ownerId = rawRecord.getRequired("st_owner_id").toLong,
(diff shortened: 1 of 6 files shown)


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
```diff
@@ -16,6 +16,28 @@ object OwnerTransformations {
...
+    // format is: "Division <N>: <some-description>"
+    val censusDivisionPattern: Regex = "(Division \\d+:)"
+    
+    /**
+      * Extract the Division value from the full text string.
+      *
+      * @param divisionString the text string to be parsed
+      * @return an integer of the division number
+    **/
+    def getCensusDivision(divisionString: String): (integer) = {
```
danxmoran, line 30: Invalid type annotation here. Once you fix it the body will need tweaking to return the right type (it's currently a string)

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
```diff
@@ -44,7 +66,12 @@ object OwnerTransformations {
       ocHouseholdChildCount = rawRecord.getOptionalNumber("oc_children_household"),
       ssHouseholdDogCount = rawRecord.getOptionalNumber("ss_num_dogs_hh"),
       ocPrimaryResidenceState = rawRecord.getOptional("oc_address1_state"),
-      ocPrimaryResidenceCensusDivision = rawRecord.getOptional("oc_address1_division"),
+      ocPrimaryResidenceCensusDivision = getCensusDivision(rawRecord.getOptional("oc_address1_division")),
+
+            ocSecondaryResidenceState = secondaryAddress.flatMap {
```
danxmoran, line 71: This is already defined below, accidental copy-paste?

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/OwnerTransformationsSpec.scala
```diff
@@ -59,7 +59,7 @@ class OwnerTransformationsSpec extends AnyFlatSpec with Matchers {
     output.ssHouseholdDogCount shouldBe Some(2)
     // residence fields
     output.ocPrimaryResidenceState shouldBe Some("OH")
-    output.ocPrimaryResidenceCensusDivision shouldBe Some("Division 3: East North Central")
+    output.ocPrimaryResidenceCensusDivision shouldBe Some(3)
```
danxmoran, line 62: Should this be 33? If we're extracting 3, we're losing the 2nd digit.

It looks like the other test is checking the double-digit case, so I'd actually say we should revert the source data in this one to be 3.

raaidbroad: Seconded; looks like it should be 33

PR #31, opened 2020-06-16
Ra dspdc 879 infectious disease

raaidbroad:
This starts off the hles_health_condition work. I added the schema.json for it, initialized the HealthTransformations piece, and wrote tests for it. This **only** looks at infectious disease pieces, so there are plenty of gaps to fill in right now for future tickets. Since this is kicking this off, please take a look at how I've decided to organize it. I tried to reduce hardcoding things where I could, but I think the lists of prefixes are pretty hard to get around. Any additional tests that might be helpful to write would also be welcome suggestions.

### DIFF
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
@@ -49,9 +49,9 @@ object HealthTransformations {
         case ("infect_other", categorical) =>
           createHealthConditionRow(
             rawRecord,
-            s"dx_infect_other",
-            1L,
-            categorical.toLong,
+            "dx_infect_other",
+            infectiousDiseaseCondition,
+            categorical,
             isCongenital = false,
             rawRecord.getOptional("hs_dx_infect_other_spec")
           )
@@ -60,15 +60,19 @@ object HealthTransformations {
           createHealthConditionRow(
             rawRecord,
             s"dx_${disease}",
-            1L,
-            categorical.toLong
+            infectiousDiseaseCondition,
+            categorical
           )
       }
     } else {
       None
     }
 
-  val infectious_diseases: Map[String, Int] = Map(
+  // list conditions and assign categorical Longs
+  val infectiousDiseaseCondition = 1L
+
+  // specific condition type maps
+  val infectious_diseases: Map[String, Long] = Map(
     "anaplasmosis" -> 0,
     "asperg" -> 1,
     "babesio" -> 2,
@@ -109,6 +113,6 @@ object HealthTransformations {
     "toxop" -> 37,
     "tular" -> 38,
     "whpworm" -> 39,
-    "infect_other" -> 99
+    "infect_other" -> 98
   )
 }
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
@@ -4,48 +4,118 @@ import org.broadinstitute.monster.dogaging.jadeschema.table.HlesHealthCondition
 
 object HealthTransformations {
 
-  val infectious_diseases: List[(String, Int)] = List("anaplasmosis", "asperg", "babesio", "blastomy", "bordetella", "brucellosis", "campylo",
-    "chagas", "ccdia", "ccdio", "crypto", "dermato", "dstmp", "ehrlich", "fever", "gp", "giar", "granu", "hrtworm",
-    "histo", "hepato", "hkworm", "influ", "isosp", "leish", "lepto", "lyme", "mrsa", "mycob", "parvo", "plague",
-    "pythium", "rmsf", "rndworm", "slmosis", "slmpois", "tpworm", "toxop", "tular", "whpworm", "infect_other")
-    .zipWithIndex
-
   /** Parse all health-condition-related fields out of a raw RedCap record. */
-  def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] = {
+  def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
     // condition type is congenital vs infectious vs tox vs trauma vs etc
-    List("congenital", "dx_infectious", "dx_tox", "dx_trauma", "dx_eye", "dx_ear", "dx_oral",
-      "dx_skin", "dx_cardiac", "dx_respire", "dx_gi", "dx_liver", "dx_kidney", "dx_repro", "dx_ortho", "dx_neuro",
-      "dx_endo", "dx_hema", "dx_immune")
-      .zipWithIndex
-      .filter { case (condition, _) => rawRecord.getBoolean(s"hs_${condition}_yn") }
-      .flatMap {
-        // congenital conditions
-        case ("congenital", conditionCategorical) =>
-          ???
-        // infectious disease
-        case ("dx_infectious", conditionCategorical) =>
-          // process infectious disease data
-          infectious_diseases
-            .filter { case (infection, _) => rawRecord.getBoolean(s"hs_dx_${infection}") }
-            .map { case (infection, infectionCategorical) =>
-              HlesHealthCondition(
-                dogId = rawRecord.getRequired("study_id").toLong,
-                hsConditionType = conditionCategorical.toLong,
-                hsCondition = infectionCategorical.toLong,
-                // should only be populated if hs_dx_infect_other is true
-                hsConditionOtherDescription = rawRecord.getOptional("hs_dx_infect_other_spec"),
-                // infectious diseases are not categorized as congenital
-                hsConditionIsCongenital = false,
-                // no relevant field for infectious diseases
-                hsConditionCause = None,
-                // no relevant field for infectious diseases
-                hsConditionCauseOtherDescription = None,
-                hsDiagnosisYear = rawRecord.getOptionalNumber(s"hs_dx_${infection}_year"),
-                hsDiagnosisMonth = rawRecord.getOptionalNumber(s"hs_dx_${infection}_month"),
-                hsRequiredSurgeryOrHospitalization = rawRecord.getOptionalNumber(s"hs_dx_${infection}_surg"),
-                hsFollowUpOngoing = rawRecord.getOptionalBoolean(s"hs_dx_${infection}_fu")
-              )
-            }
-      }
-  }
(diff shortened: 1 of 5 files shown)


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,124 @@
...
+            hsConditionCauseOtherDescription = None,
+            hsDiagnosisYear = Some(2020),
+            hsDiagnosisMonth = Some(3),
+            hsRequiredSurgeryOrHospitalization = Some(1),
+            hsFollowUpOngoing = Some(true)
+          )
+        )
+    }
+
+  val health_conditions: List[(String, Int)] = List(
```
raaidbroad, line 59: for both these and for the infectious diseases, I am laying them out in the order they appear on RedCap in the data dictionary and then generating a categorical integer for them by their indexed position. I'm also open to explicitly writing out the categorical integer each one maps to; the only benefit I'd see there is that we, the developers, could clearly read the number. Otherwise, so long as we maintain the order here, it should operate the same.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,124 @@
...
+              hsDiagnosisYear = rawRecord.getOptionalNumber(s"hs_dx_${infection}_year"),
+              hsDiagnosisMonth = rawRecord.getOptionalNumber(s"hs_dx_${infection}_month"),
+              hsRequiredSurgeryOrHospitalization =
+                rawRecord.getOptionalNumber(s"hs_dx_${infection}_surg"),
+              hsFollowUpOngoing = rawRecord.getOptionalBoolean(s"hs_dx_${infection}_fu")
+            )
+        }
+      // an "else" case to please the compiler gods for now, will eventually be replaced by all the other health
+      // condition cases
+      case (_, conditionCategorical) =>
```
danxmoran, line 41: I don't think this pattern is scalable (there will be much repeated code across cases, and the file will be looooong). IMO it'd be better to have:
1. A generic, parameterized method that handles the actual `RawRecord` => `Option[HlesHealthCondition]` conversion for a specific condition
2. A method per each condition_type which handles the top-level Y/N filter, looping over each specific condition (delegating to the generic method), and adding any edge-cases (i.e. the "other" spec case)

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,124 @@
...
+        infectious_diseases.filter {
+          case (infection, _) => rawRecord.getBoolean(s"hs_dx_${infection}")
+        }.map {
+          case (infection, infectionCategorical) =>
+            HlesHealthCondition(
+              dogId = rawRecord.getRequired("study_id").toLong,
+              hsConditionType = conditionCategorical.toLong,
+              hsCondition = infectionCategorical.toLong,
+              // should only be populated if hs_dx_infect_other is the current infectionCategorical (value is 40)
+              hsConditionOtherDescription = if (infectionCategorical == 40) {
```
danxmoran, line 23: We shouldn't need to have this "magic" dependency between constants. We can tag each of the field names with a field for other / not other, or we could leave the "other" case out of the list and have a special one-off call for it.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,124 @@
...
+    "rmsf",
+    "rndworm",
+    "slmosis",
+    "slmpois",
+    "tpworm",
+    "toxop",
+    "tular",
+    "whpworm",
+    "infect_other"
+  ).zipWithIndex
```
danxmoran, line 123: `zipWithIndex` worries me, it'll be super easy to accidentally change outputs by rearranging the list. It's annoying, but would we instead have a `Map[(String, Int)]` and explicitly list out the numeric values?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,124 @@
...
+    "dx_gi",
+    "dx_liver",
+    "dx_kidney",
+    "dx_repro",
+    "dx_ortho",
+    "dx_neuro",
+    "dx_endo",
+    "dx_hema",
+    "dx_immune"
+  ).zipWithIndex
```
danxmoran, line 79: I'm not sure if we need this mapping based on my proposed refactor above.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,114 @@
...
+          hsFollowUpOngoing = rawRecord.getOptionalBoolean(s"hs_${conditionName}_fu")
+        )
+      )
+    } else {
+      None
+    }
+
+  /** Parse all infectious disease related fields out of a raw RedCap record. */
+  def mapInfectiousDisease(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
+    if (rawRecord.getBoolean("hs_dx_infectious_yn")) {
```
danxmoran, line 45: I think this will need to use `getOptionalBoolean`

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,114 @@
...
+      // iterate over all infectious diseases
+      infectious_diseases.flatMap {
+        // "other" case
+        case ("infect_other", categorical) =>
+          createHealthConditionRow(
+            rawRecord,
+            s"dx_infect_other",
```
danxmoran, line 52:
```suggestion
            "dx_infect_other",
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,114 @@
...
+            1L,
```
danxmoran, line 53: Worth making this a top-level named constant, since it's used in 2 places?

I'm not sure if this will stay "1" over time, DAP will probably expect the numbering to match the order of questioning in HLES. Having a group of constants will make it easier for us to be sure there are no duplicates & the ordering is what we'd expect.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,114 @@
...
+            categorical.toLong,
```
danxmoran, line 54: Would be better to have the map values be `Long` to start. The compiler might complain if you don't add the `L` suffix to each number.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -0,0 +1,114 @@
...
+    "pythium" -> 31,
+    "rmsf" -> 32,
+    "rndworm" -> 33,
+    "slmosis" -> 34,
+    "slmpois" -> 35,
+    "tpworm" -> 36,
+    "toxop" -> 37,
+    "tular" -> 38,
+    "whpworm" -> 39,
+    "infect_other" -> 99
```
danxmoran, line 112: 98 is used elsewhere as the "other" marker (99 is usually used for "unknown")

danxmoran:
Don't worry about codecov, there's a ticket in the next sprint to configure the %s we actually want to aim for

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: If I'm reading the data dictionary correctly, this should always be present; it doesn't have a "Show the field ONLY if <condition>" listed

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: So like a constant `map` of all the top level conditions, referenced where appropriate?

danxmoran: I was thinking constants like:
```scala
val EyeCondition = 1L
// moar
val InfectiousDisease = 10L
```

danxmoran: The map could also work, but then you end up with repeated magic strings at the lookup points

raaidbroad: lol we're gonna have so many! But ye, I'm down for just listing constants instead of a map

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
danxmoran: My thinking for the goal is: We should be able to call the top-level `mapHealthConditions` method on `RawRecord(1, Map.empty)` and get out `HlesHealthCondition.init(1)`. As long as we can do that w/ this as `getBoolean` instead of `getOptionalBoolean`, I'm ok with it.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
danxmoran: It might be worth making a "constants" object in a separate file to store all the magic numbers & maps. Normally I don't like it, but here it feels potentially justified

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: I'll add that as a test

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: ^I'll leave the constants in here, and if/when someone feels it is getting out of control then we can move it to a separate constants object in another file.

PR #33, opened 2020-06-17
Ra dspdc 863 health eyes

raaidbroad:
Add eye condition pieces; let me know if y'all can think of a better/cleaner way to do logic stuff for this, particularly on the blind case. I think the tests are adequate but also let me know if there are more things I ought to test for.

### DIFF
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
@@ -6,7 +6,11 @@ object HealthTransformations {
 
   /** Parse all health-condition-related fields out of a raw RedCap record. */
   def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    Iterable.concat(mapInfectiousDisease(rawRecord), mapEyeDisease(rawRecord))
+    Iterable.concat(
+      mapInfectiousDisease(rawRecord),
+      mapEyeDisease(rawRecord),
+      mapCongenitalEyeDisorder(rawRecord)
+    )
 
   /** Generic helper method for creating Hles Health Condition rows. */
   def createHealthConditionRow(
@@ -63,6 +67,31 @@ object HealthTransformations {
       }
     } else None
 
+  def mapCongenitalEyeDisorder(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
+    if (rawRecord.getBoolean("hs_congenital_yn") && rawRecord.getBoolean("hs_cg_eye_disorders_yn"))
+      eye_cg_disorders.flatMap {
+        // "other" case
+        case ("other", categorical) =>
+          createHealthConditionRow(
+            rawRecord,
+            "cg_eye_other",
+            eyeDiseaseCondition,
+            categorical,
+            isCongenital = true,
+            conditionOtherDescription = rawRecord.getOptional("hs_cg_eye_other_spec")
+          )
+        // general case
+        case (disorder, categorical) =>
+          createHealthConditionRow(
+            rawRecord,
+            s"cg_eye_${disorder}",
+            eyeDiseaseCondition,
+            categorical,
+            isCongenital = true
+          )
+      }
+    else None
+
   def mapEyeDisease(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
     if (rawRecord.getBoolean("hs_dx_eye_yn")) {
       eye_diseases.flatMap {
@@ -96,11 +125,49 @@ object HealthTransformations {
       }
     } else None
 
-  // list conditions and assign categorical Longs
+  ////////////////////////////////
+  //// condition categoricals ////
+  ////////////////////////////////
   val eyeDiseaseCondition = 1L
   val infectiousDiseaseCondition = 50L
 
-  // specific condition type maps
+  //////////////////////////////////////
+  //// specific condition type maps ////
+  //////////////////////////////////////
+
+  // eye disorders and diseases
+  val eye_cg_disorders: Map[String, Long] = Map(
+    "blind" -> 0,
+    "cat" -> 1,
+    "glauc" -> 2,
+    "kcs" -> 3,
+    "ppm" -> 4,
+    "miss" -> 5,
+    "other" -> 98
+  )
+
+  val eye_diseases: Map[String, Long] = Map(
+    "cat" -> 0,
+    "blind" -> 1,
+    "ce" -> 2,
+    "conj" -> 3,
+    "cu" -> 4,
+    "dist" -> 5,
+    "kcs" -> 6,
+    "ectrop" -> 7,
+    "entrop" -> 8,
+    "glauc" -> 9,
+    "ilp" -> 10,
+    "ic" -> 11,
+    "jcat" -> 12,
+    "ns" -> 13,
+    "pu" -> 14,
+    "pra" -> 15,
+    "rd" -> 16,
+    "uvei" -> 17,
+    "eye_other" -> 98
+  )
+
   val infectious_diseases: Map[String, Long] = Map(
     "anaplasmosis" -> 0,
     "asperg" -> 1,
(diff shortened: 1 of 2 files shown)


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -6,7 +6,7 @@ object HealthTransformations {
 
   /** Parse all health-condition-related fields out of a raw RedCap record. */
   def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    mapInfectiousDisease(rawRecord)
+    mapInfectiousDisease(rawRecord) ++ mapEyeDisease(rawRecord)
```
danxmoran, line 9: Would use `Iterable.concat` instead of `++` to set the pattern moving forward

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -68,8 +67,42 @@ object HealthTransformations {
       None
     }
 
+  def mapEyeDisease(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
+    if (rawRecord.getBoolean("hs_dx_eye_yn")) {
```
danxmoran, line 71: What about the congenital cases?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -68,8 +67,42 @@ object HealthTransformations {
...
+          )
+        }
+        case (disease, categorical) =>
+          createHealthConditionRow(rawRecord, s"dx_${disease}", eyeDiseaseCondition, categorical)
+      }
+    } else { None }
+
   // list conditions and assign categorical Longs
   val infectiousDiseaseCondition = 1L
+  val eyeDiseaseCondition = 2L
```
danxmoran, line 105: According to the RedCap layout, I think eyes should have value 1 (they come first in the ordering of questions)

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: There are congenital cases?! WHOOPS

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: Where is the layout? I was going by order of stuff [here](https://redcap.fredhutch.org/redcap_v9.8.2/Design/data_dictionary_codebook.php?pid=895), where hs_dx_infectious_yn is at number 1275 but hs_dx_eye_yn is at 1623

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: Oh my goodness I didn't realize there was a hs_cg_eye_disorders_yn

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
danxmoran: The congenital things are asked first, so I've been thinking we'd do a partial-overlay style of ordering: Assign indexes based on the CG ordering, then add the diagnosed-only conditions to the end.

raaidbroad: Makes sense; I am still coming down from the shock of learning that there are completely differently named congenital eye things =P

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -68,8 +67,42 @@ object HealthTransformations {
...
+            rawRecord,
+            "dx_eye_other",
+            eyeDiseaseCondition,
+            categorical,
+            conditionOtherDescription = rawRecord.getOptional("hs_dx_eye_other_spec")
+          )
+        case ("blind", categorical) => {
+          val isCauseKnown = rawRecord.getBoolean("hs_dx_eye_cause_yn")
+          val conditionCause = if (isCauseKnown) { rawRecord.getOptionalNumber("hs_dx_eye_cause") }
+          else { None }
```
danxmoran, line 85: Wot is dis scalafmtery

Better would be either:
```scala
val conditionCause =
  if (isCauseKnown) rawRecord.getOptionalNumber("hs_dx_eye_cause") else None
```
Or
```scala
val conditionCause = if (isCauseKnown) {
  rawRecord.getOptionalNumber("hs_dx_eye_cause")
} else {
  None
}
```
Same for the other if/else clauses in this file, can you update them?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -64,14 +65,109 @@ object HealthTransformations {
...
       }
-    } else {
-      None
-    }
+    } else None
 
-  // list conditions and assign categorical Longs
-  val infectiousDiseaseCondition = 1L
+  def mapCongenitalEyeDisorder(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
+    if (rawRecord.getBoolean("hs_congenital_yn") && rawRecord.getBoolean("hs_cg_eye_disorders_yn"))
```
quazi-broad, line 71: does the hs_congenital_yn flag show up for non eye congential disorders?

raaidbroad: Yes

raaidbroad:
> LGTM. I think this is one of the first times where the Spec has way more lines than the actual transformation code. Is that cause there are so many different use cases/scenarios you had to write out? On that note, did you just try to think through the different unit test cases or were you able to look at actual data to help model those?

I wanted to be thorough for the different possibilities of blindness; it has the most possibilities out of any conditions I think, and it seems like a good idea to check every possible case for it.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -64,14 +65,109 @@ object HealthTransformations {
...
+
+  // eye disorders and diseases
+  val eye_cg_disorders: Map[String, Long] = Map(
+    "blind" -> 0,
+    "cat" -> 1,
+    "glauc" -> 2,
+    "kcs" -> 3,
+    "ppm" -> 4,
+    "miss" -> 5,
+    "other" -> 98
```
snwessel, line 146: Only asking for my own understanding - Are these values based on the order of the Redcap questions? Or is this mapping defined somewhere that you're copying it over from?

raaidbroad: Based on order. We have to generate/define numerical categoricals for these conditions, and instead of like iterating over a list and using the index, we want to write them out so that we have some clearly defined human-readable source of truth for what the number for a given condition is.

snwessel: Great! Thanks, that makes sense.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/HealthTransformationsSpec.scala
```diff
@@ -42,13 +43,122 @@ class HealthTransformationsSpec extends AnyFlatSpec with Matchers {
     "hs_dx_infect_other_fu" -> Array("1")
   )
 
-  it should "correctly map infectious disease values when values are defined for a single disease" in {
-    val exampleInfectiousDiseaseRecord = RawRecord(id = 1, singleInfectiousDiseaseFields)
+  private val singleEyeDisease = Map[String, Array[String]](
```
snwessel, line 46: It looks like these input values are each only used in one test... would it make sense to move them into (or nearer to) the tests where they're used to make it easier to compare the input and output? I'm mostly suggesting that because this file will get pretty big as we add more health conditions.

raaidbroad: This is a good point! I have another PR coming in right after this; for the sake of getting this one through, I will move the input values around in the next PR.

PR #35, opened 2020-06-18
Sw add tsv conversion script

snwessel:
I had written this python script as a part of an earlier ticket to upload transformed Dog Aging data to Terra. Since we'll probably need to do that again in the near future, I'm adding the script here so others can use it. 

It's a bit messy, but I didn't want to spend too much time on it. Let me know if there's anything I should add or clean up!

### DIFF
--- a/hack/convert-output-to-tsv.py
+++ b/hack/convert-output-to-tsv.py
@@ -1,5 +1,3 @@
-# A script to convert transformed data into TSV files
-
 import json
 import csv
 from os import listdir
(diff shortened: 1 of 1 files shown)


hack/convert-output-to-tsv.py
```diff
@@ -0,0 +1,42 @@
+# A script to convert transformed data into TSV files
+# Note: Before uploading to Terra, manually go in and update PK column name to be prefixed by "entity:"
```
snwessel, line 2: I did this manually since I didn't want to mess with all of the JSON dictionaries. I'd be happy to add it to the script though if anyone prefers that.

danxmoran: If it isn't a huge headache, I think it'd be worth adding

hack/convert-output-to-tsv.py
```diff
@@ -0,0 +1,42 @@
...
+
+import json
+import csv
+from os import listdir
+import sys
+
+if (len(sys.argv) < 3):
+    print "Please provide the input directory and output directory as arguments!"
+
+table_names = ["hles_dog", "hles_owner"]
```
danxmoran, line 12: Does this need to be hard-coded, or could we just `ls` the output / tables directory?

snwessel: Good point. I'll have it check the directory 👍

hack/convert-output-to-tsv.py
```diff
@@ -0,0 +1,51 @@
...
+    primary_key_list = []
+    with open(schema_dir + table_name + '.table.json') as schema_file:
+        schema = json.load(schema_file)
+        for column in schema['columns']:
+            column_name = column['name']
+            if (column.get('type') == 'primary_key'):
+                primary_key_list.append(column_name)
+                column_list.append(pk_prefix + column_name)
+            else: 
+                column_list.append(column_name)
```
danxmoran, line 29: This loop isn't going to work any more now that we've scattered columns across multiple files.

Instead of collecting column names from the schemas, could we build up a `set()` of keys found while adding data to `obj_list` below?

snwessel: Oh, true... And yeah, that sounds like a good plan.

PR #37, opened 2020-06-22
QH dspdc 1138 transform dog routine environment

quazi-broad:
** Added transformation logic for routine environment fields.
** Updated json schema - missing one field.
Need to follow up on "FIXME" notes, had a few questions.
Also need to write unit tests.


quazi-broad:
Code doesn't compile atm, just wanted to get all the work I've already done up in the draft.

danxmoran:
@quazi-broad formatting is failing, so I can't see the compile & test output. Can you fix & re-push before I review?

quazi-broad:
Sorry for the delay, it took a lot longer to resolve the errors from the changes I made.
Code should now compile and unit tests have been expanded to cover every single case.
I left a bunch of "FIXME" comments in the code - 2 in the actual transformation code and a bunch in the unit tests.
Going to go through and clean those up before I merge, otherwise please have at it!

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,370 @@
...
+        deRecentToxinsOrHazardsIngestedOther = recentToxinsOrHazardsIngestedOther,
+        //de_ingest_bad is deprecated, check it first, then check de_ingest_bad_what_other
+        deRecentToxinsOrHazardsIngestedOtherDescription =
+          if (toxinsDescription.exists(_.trim.nonEmpty)) toxinsDescription
+          // ONLY if: de_ingest_bad_what(98) = 1
+          else if (recentToxinsOrHazardsIngestedOther.get)
+            rawRecord.getOptional("de_ingest_bad_what_other")
+          else None,
+        deRecentToxinsOrHazardsIngestedRequiredVet =
+          rawRecord.getOptionalBoolean("de_ingest_bad_er_yn")
```
quazi-broad, line 305: Any suggestions on cleaning up/refactoring this to be more readable?
This particular piece was pretty annoying.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformationsSpec.scala
```diff
@@ -0,0 +1,786 @@
...
+      RoutineEnvironmentTransformations.mapRoutineEnvironment(RawRecord(1, example3))
+
+    output3.deRoutineToys.value shouldBe false
+    output3.deLicksChewsOrPlaysWithNonToys.value shouldBe true
+
+    // de_toys_yn = '0', de_chew_other_yn = '0'
+    val example4 = Map[String, Array[String]](
+      "de_toys_yn" -> Array("0"),
+      "de_chew_other_yn" -> Array("0")
+    )
```
quazi-broad, line 469: I feel like this test case is not that useful, should I remove it?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,370 @@
...
+        deDogparkDaysPerMonth = rawRecord.getOptionalNumber("de_dogpark_freq"),
+        deDogparkTravelWalk = dogparkTravel.map(_.contains("1")),
+        deDogparkTravelDrive = dogparkTravel.map(_.contains("2")),
+        deDogparkTravelBike = dogparkTravel.map(_.contains("3")),
+        deDogparkTravelPublicTransportation = dogparkTravel.map(_.contains("4")),
+        deDogparkTravelOther = dogparkTravelOther,
+        //ONLY if: de_dogpark_get_to(98) = '1'
+        deDogparkTravelOtherDescription =
+          if (dogparkTravelOther.contains(true)) rawRecord.getOptional("de_dogpark_get_to_other")
+          else None,
```
danxmoran, line 52: Style nitpick: Can you wrap the branches in curly brackets for the if/elses that spill to multiple lines?
```scala
if (thing) {
  // stuff
} else {
  // other stuff
}
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,370 @@
...
+        // Compute using hour & min columns, only if de_dogpark_yn = Yes
+        // Multiply hours by 60 to get minutes
+        deDogparkTravelTimeMinutes = Some(
+          (rawRecord.getRequired("de_dogpark_hr").toLong * 60)
+            + rawRecord.getRequired("de_dogpark_min").toLong
+        )
```
danxmoran, line 58: If you want to avoid `getRequired` only to re-wrap in a `Some`, you could use a `for` comprehension in these places to keep things in the `Option` context:
```scala
for {
  hours <- rawRecord.getOptionalNumber("de_dogpark_hr")
  minutes <- rawRecord.getOptionalNumber("de_dogpark_min") 
} yield {
  hours * 60 + minutes
}
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
danxmoran: Same for all the conditionals below

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
danxmoran: Same for the other cases that match this pattern

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
danxmoran: I think something like this would work:
```scala
recentToxinsOrHazardsIngestedOther.fold(toxinsDescription) {
  if (_) rawRecord.getOptional("de_ingest_bad_what_other") else None
}
```
And I'd tweak the definition of `toxinsDescription` to include the trim & filter:
```scala
val toxinsDescription = rawRecord.getOptional("de_ingest_bad").map(_.trim).filter(_.nonEmpty)
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,414 @@
...
+        //de_ingest_bad is deprecated, check it first, then check de_ingest_bad_what_other
+        deRecentToxinsOrHazardsIngestedOtherDescription =
+          //  if (toxinsDescription.exists(_.trim.nonEmpty)) {
+          //    toxinsDescription
+          //    // ONLY if: de_ingest_bad_what(98) = 1
+          //  } else if (recentToxinsOrHazardsIngestedOther.get) {
+          //    rawRecord.getOptional("de_ingest_bad_what_other")
+          //  } else None,
+          // ONLY if: de_ingest_bad_what(98) = 1
+          recentToxinsOrHazardsIngestedOther.fold(toxinsDescription) {
```
quazi-broad, line 343: I implemented those changes but it's failing one of my unit test cases where:
"de_ingest_bad" (toxinsDescription) is provided and "de_ingest_bad_what_other" is not defined.
I get the error: 
The Option on which value was invoked was not defined. (RoutineEnvironmentTransformationsSpec.scala:559)

In that case,
"de_ingest_bad" -> Array("Some nasty stuff"),
"de_ingest_bad_what" -> Array("1", "2", "3", "4", "5"), (OTHER option not indicated)

If I am understanding the logic correctly, we should be checking toxinsDescription (de_ingest_bad) first and if that isn't available, pull "de_ingest_bad_what_other" (but that field only exists when de_ingest_bad_what(98) = 1).

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformationsSpec.scala
danxmoran: If you think we have good enough coverage without it, sure

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
danxmoran: Ah, I see the issue. Your test examples include both "de_ingest_bad" and "de_ingest_bad_what", but this code is written to assume that only one or the other will ever be present. I believe the assumption is correct: when they introduced de_ingest_bad_what, they removed de_ingest_bad, so we should never see the two in a single record.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,414 @@
...
+        deDogparkTravelBike = dogparkTravel.map(_.contains("3")),
+        deDogparkTravelPublicTransportation = dogparkTravel.map(_.contains("4")),
+        deDogparkTravelOther = dogparkTravelOther,
+        //ONLY if: de_dogpark_get_to(98) = '1'
+        deDogparkTravelOtherDescription = if (dogparkTravelOther.contains(true)) {
+          rawRecord.getOptional("de_dogpark_get_to_other")
+        } else None,
+        // Compute using hour & min columns, Multiply hours by 60 to get minutes
+        deDogparkTravelTimeMinutes = for {
+          hours <- rawRecord.getOptionalNumber("de_dogpark_hr")
```
snwessel, line 67: In other places, we changed the schema to store hour and minute fields separately because the "hour" fields don't allow the user to enter an integer themselves, they have to select from 0, 1, 2, ... 11, or "12+" (and it doesn't make sense to convert an option like 12+ into minutes).

I'm not sure who would travel 12+ hours to go to the dog park anyway 🤷‍♀️ but maybe it would be worth changing for the sake of consistency? What do others think?

danxmoran: Ergh, fair point... It makes me sad but it makes sense

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,414 @@
...
+        deRecreationalSpacesTravelBike = dogRecSpacesTravel.map(_.contains("3")),
+        deRecreationalSpacesTravelPublicTransportation = dogRecSpacesTravel.map(_.contains("4")),
+        deRecreationalSpacesTravelOther = dogRecSpacesTravelOther,
+        //ONLY if: de_spaces_get_to(98) = '1'
+        deRecreationalSpacesTravelOtherDescription = if (dogRecSpacesTravelOther.contains(true)) {
+          rawRecord.getOptional("de_spaces_get_to_other")
+        } else None,
+        // Compute using hour & min columns, Multiply hours by 60 to get minutes
+        deRecreationalSpacesTravelTimeMinutes = for {
+          hours <- rawRecord.getOptionalNumber("de_spaces_hr")
```
snwessel, line 108: Same comment about hours and minutes here. They should probably be split apart in the schema since the hours value comes from a dropdown.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
quazi-broad: Thanks for the suggestion, Sarah! I expanded the schema to add separate "hours" fields for these cases. I removed the code that was adding the minutes and hours.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
quazi-broad: That makes sense! I made sure all the test data had either the deprecated field OR the new one. I just had to make one change, making toxinsDescription the default (instead of None) within that function and now it passes all the unit tests

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,362 @@
...
+        deDogparkDaysPerMonth = rawRecord.getOptionalNumber("de_dogpark_freq"),
+        deDogparkTravelWalk = dogparkTravel.map(_.contains("1")),
+        deDogparkTravelDrive = dogparkTravel.map(_.contains("2")),
+        deDogparkTravelBike = dogparkTravel.map(_.contains("3")),
+        deDogparkTravelPublicTransportation = dogparkTravel.map(_.contains("4")),
+        deDogparkTravelOther = dogparkTravelOther,
+        //ONLY if: de_dogpark_get_to(98) = '1'
+        deDogparkTravelOtherDescription = if (dogparkTravelOther.contains(true)) {
+          rawRecord.getOptional("de_dogpark_get_to_other")
+        } else None,
```
danxmoran, line 52: `else` branches still need brackets:
```scala
else {
  None
}
```
Across the board

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
```diff
@@ -0,0 +1,362 @@
...
+        deSitterOrDaycareTravelWalk = dogSitterTravel.map(_.contains("1")),
+        deSitterOrDaycareTravelDrive = dogSitterTravel.map(_.contains("2")),
+        deSitterOrDaycareTravelBike = dogSitterTravel.map(_.contains("3")),
+        deSitterOrDaycareTravelPublicTransportation = dogSitterTravel.map(_.contains("4")),
+        deSitterOrDaycareTravelOther = dogSitterTravelOther,
+        // ONLY if: de_sitter_how(98) = '1'
+        deSitterOrDaycareTravelOtherDescription = if (dogSitterTravelOther.contains(true)) {
+          rawRecord.getOptional("de_sitter_how_other")
+        } else None,
+        //Compute using hour & min columns, Multiply hours by 60 to get minutes
```
danxmoran, line 143: Outdated comment

quazi-broad:
Expanded the else blocks and removed that old comment. Thanks!

PR #38, opened 2020-06-23
Ra dspdc 1141 transformation refactor

raaidbroad:
Restructure how we're doing the health condition transformation. The existing tests still pass, and this way it seems like we won't have to write any additional business logic and all we'll have to do is add the rest of the conditions. I am SURE that the `processHealthCondition` function can be cleaned up/modularized further, so let me hear it if you think it can. I am also quite open to putting the massive Maps/List at the bottom into an object in a separate file, and also open to this whole endeavor not being the way we'd want to do this moving forward.

### DIFF
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
@@ -10,10 +10,13 @@ package org.broadinstitute.monster.dap
   * @param nonCongenital a tuple where the first element is the piece of the target field name that fits in "hs_*",
   *                      like "dx_blind", and where the second element is the higher level boolean flag name that fits
   *                      in "hs_*_yn", like "dx_eye", that the first depends on
+  * @param isOther       a boolean flag that indicates this condition is an "other" condition, helpful for the business
+  *                      logic to correctly transform data
   */
 case class HealthCondition(
   conditionType: String,
   condition: String,
   congenital: Option[(String, String)] = None,
-  nonCongenital: Option[(String, String)] = None
+  nonCongenital: Option[(String, String)] = None,
+  isOther: Boolean = false
 )
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
@@ -1,0 +1,19 @@
+package org.broadinstitute.monster.dap
+
+/**
+  *
+  * @param conditionType a name to categorize the condition type by, like "eye", "ear", "cardiac", etc.
+  * @param condition     a specific condition within a condition type, like "blind", "cat", "glauc", "other", etc.
+  * @param congenital    a tuple where the first element is the piece of the target field name that fits in
+  *                      "hs_*_yn", like "cg_eye_blind", and where the second element is the higher level boolean
+  *                      flag name that fits in "hs_*_yn", like "cg_eye_disorders", that the first depends on
+  * @param nonCongenital a tuple where the first element is the piece of the target field name that fits in "hs_*",
+  *                      like "dx_blind", and where the second element is the higher level boolean flag name that fits
+  *                      in "hs_*_yn", like "dx_eye", that the first depends on
+  */
+case class HealthCondition(
+  conditionType: String,
+  condition: String,
+  congenital: Option[(String, String)] = None,
+  nonCongenital: Option[(String, String)] = None
+)
--- a/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
+++ b/hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
@@ -6,11 +6,7 @@ object HealthTransformations {
 
   /** Parse all health-condition-related fields out of a raw RedCap record. */
   def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    Iterable.concat(
-      mapInfectiousDisease(rawRecord),
-      mapEyeDisease(rawRecord),
-      mapCongenitalEyeDisorder(rawRecord)
-    )
+    healthConditions.flatMap(healthCondition => processHealthCondition(rawRecord, healthCondition))
 
   /** Generic helper method for creating Hles Health Condition rows. */
   def createHealthConditionRow(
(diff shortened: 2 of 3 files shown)


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -6,11 +6,7 @@ object HealthTransformations {
 
   /** Parse all health-condition-related fields out of a raw RedCap record. */
   def mapHealthConditions(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    Iterable.concat(
-      mapInfectiousDisease(rawRecord),
-      mapEyeDisease(rawRecord),
-      mapCongenitalEyeDisorder(rawRecord)
-    )
+    healthConditions.flatMap(healthCondition => processHealthCondition(rawRecord, healthCondition))
```
danxmoran, line 9: If you want:
```suggestion
    healthConditions.flatMap(processHealthCondition(rawRecord, _))
```
:scala:

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,241 @@ object HealthTransformations {
...
-  def mapCongenitalEyeDisorder(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    if (rawRecord.getBoolean("hs_congenital_yn") && rawRecord.getBoolean("hs_cg_eye_disorders_yn"))
-      eye_cg_disorders.flatMap {
-        // "other" case
-        case ("other", categorical) =>
+  def processHealthCondition(
+    rawRecord: RawRecord,
+    healthCondition: HealthCondition
+  ): Iterable[HlesHealthCondition] = {
+    val congenital: Option[HlesHealthCondition] = if (healthCondition.congenital.isDefined) {
```
danxmoran, line 45: Instead of `isDefined` here and `get` below, this is more idiomatic:
```scala
val congenital: Option[HlesHealthCondition] = healthCondition.congenital.flatMap {
  case (thing1, thing2) => ???

}
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,241 @@ object HealthTransformations {
...
-
-  def mapEyeDisease(rawRecord: RawRecord): Iterable[HlesHealthCondition] =
-    if (rawRecord.getBoolean("hs_dx_eye_yn")) {
-      eye_diseases.flatMap {
-        // "other" case
-        case ("eye_other", categorical) =>
+        else None
+      else None
+    } else None
+    val nonCongenital: Option[HlesHealthCondition] = if (healthCondition.nonCongenital.isDefined) {
```
danxmoran, line 72: Same comment as above, prefer `flatMap` to `isDefined` + `get`

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,241 @@ object HealthTransformations {
...
-  ////////////////////////////////
-  val eyeDiseaseCondition = 1L
-  val infectiousDiseaseCondition = 50L
+  val healthConditions: List[HealthCondition] = List(
+    // eye conditions
+    HealthCondition(
+      "eye",
+      "blind",
+      Some(("cg_eye_blind", "cg_eye_disorders")),
+      Some(("dx_blind", "dx_eye"))
```
danxmoran, line 121: Some of this paren-ing might look better using the `->` syntax
```suggestion
      Some("dx_blind" -> "dx_eye")
```

danxmoran: But no pressure if the way it is now is preferred

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,241 @@ object HealthTransformations {
...
+    "entrop" -> 11,
+    "ilp" -> 12,
+    "ic" -> 13,
+    "jcat" -> 14,
+    "ns" -> 15,
+    "pu" -> 16,
+    "pra" -> 17,
+    "rd" -> 18,
+    "uvei" -> 19,
+    "eye_other" -> -1,
```
danxmoran, line 234: I worry that using a negative number here will cause some bad downstream effect bc nothing works the way it should.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,241 @@ object HealthTransformations {
...
-      eye_diseases.flatMap {
-        // "other" case
-        case ("eye_other", categorical) =>
+        else None
+      else None
+    } else None
+    val nonCongenital: Option[HlesHealthCondition] = if (healthCondition.nonCongenital.isDefined) {
+      if (rawRecord.getBoolean(s"hs_${healthCondition.nonCongenital.get._2}_yn"))
+        // non-congenital other
+        if (conditions.apply(healthCondition.condition) == -1L)
```
danxmoran, line 75: I would rather not key off a magic value here. Can we include a boolean in the `HealthCondition` class to drive this instead?

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: THANK YOU, I knew this felt weird

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: lmao yeah good point. I'll set it to like 999 or something for now

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: ah so like an `isOther` flag? Sure thing! I was toying with that idea anyway, lezdoit

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
raaidbroad: I'd very much prefer to not go through and change all the commas to arrows 😅

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,248 @@ object HealthTransformations {
...
+    rawRecord: RawRecord,
+    healthCondition: HealthCondition
+  ): Iterable[HlesHealthCondition] = {
+    val congenital: Option[HlesHealthCondition] = healthCondition.congenital.flatMap {
+      case (target, dependent) =>
+        // if congenital, then check hs_congenital_yn && healthCondition.congenital._2
+        if (rawRecord.getBoolean("hs_congenital_yn") &&
+            rawRecord.getBoolean(s"hs_${dependent}_yn"))
+          // congenital other
+          if (healthCondition.isOther)
```
danxmoran, line 51: Style nit: can you add brackets around the `if` and `else` arms across the board? And make sure the bodies are pushed to newlines:
```scala
if (thing) {
 // foo
} else if (otherThing) {
  // bar
} else {
 // baz
}
```

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
```diff
@@ -0,0 +1,22 @@
+package org.broadinstitute.monster.dap
+
+/**
+  *
+  * @param conditionType a name to categorize the condition type by, like "eye", "ear", "cardiac", etc.
+  * @param condition     a specific condition within a condition type, like "blind", "cat", "glauc", "other", etc.
+  * @param congenital    a tuple where the first element is the piece of the target field name that fits in
+  *                      "hs_*_yn", like "cg_eye_blind", and where the second element is the higher level boolean
+  *                      flag name that fits in "hs_*_yn", like "cg_eye_disorders", that the first depends on
```
danxmoran, line 9: My first assumption is that the order would be flipped (gate / dependent first), but I'll defer to the rest of the team on what's most clear.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,248 @@ object HealthTransformations {
...
+            createHealthConditionRow(
+              rawRecord,
+              healthCondition.congenital.get._1,
```
danxmoran, line 54: Lurking .get

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -42,173 +38,248 @@ object HealthTransformations {
...
+            val conditionCause =
+              if (isCauseKnown) rawRecord.getOptionalNumber("hs_dx_eye_cause") else None
+            createHealthConditionRow(
+              rawRecord,
+              target,
+              conditionTypes.apply(healthCondition.conditionType),
+              conditions.apply(healthCondition.condition),
+              conditionCause = conditionCause,
+              conditionCauseOtherDescription =
+                if (isCauseKnown && (conditionCause.getOrElse(false) == 98))
```
danxmoran, line 96: `getOrElse(false)` and `98` never make sense to compare, one side or the other needs updated?

raaidbroad: ahaha yup ty

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthCondition.scala
snwessel: I don't have a strong preference either way.

PR #47, opened 2020-07-01
Sw dspdc 867 map cardiac health conditions

snwessel:
I made a couple of changes to the transformation method handle a field that doesn't follow the normal pattern:

The `hs_dx_cardiac_valve` field is a description field for valve disorders, similar to the `other_spec` fields we normally see. If it followed the normal naming pattern (which it doesn't), we'd expect it to be called `hs_dx_cardiac_vd_spec`. 

I added a `descriptionSuffixOverride` arg to HealthCondition to use in cases like this - feel free to give feedback on it.


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
```diff
@@ -72,7 +75,10 @@ object HealthTransformations {
             }
           )
         } else if (healthCondition.isOther) {
-          base.copy(hsConditionOtherDescription = rawRecord.getOptional(s"${prefix}_spec"))
+          val descriptionFieldName = healthCondition.descriptionSuffixOverride
+            .map(suffix => s"${dxKey.dataPrefix}_$suffix")
+            .getOrElse(s"${prefix}_spec")
+          base.copy(hsConditionOtherDescription = rawRecord.getOptional(descriptionFieldName))
```
snwessel, line 81: This is the same change as above. If anyone has thoughts on how to reduce the duplicate code, let me know.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/healthcondition/HealthCondition.scala
```diff
@@ -21,7 +23,8 @@ sealed abstract class HealthCondition(
   val both: Option[String] = None,
   val cg: Option[String] = None,
   val dx: Option[String] = None,
-  val isOther: Boolean = false
+  val isOther: Boolean = false,
+  val descriptionSuffixOverride: Option[String] = None
```
danxmoran, line 27: Good instinct, but a potential refinement: instead of `isOther` and `descriptionSuffixOverride`, could we just have a `descriptionSuffix`? It'd default to `None`, but the "other" cases would set it to "other", and this new weird case would set it to "valve"

snwessel: Yeah, I had been thinking about that. It would mean setting the descriptionSuffix to  "other_spec" for the 20 or so normal description fields, but is definitely easier to read avoids making the transformation more complex.

snwessel: @danxmoran I looked into this some more and found a couple of cases that make it difficult (or maybe impossible?) to do with a single `descriptionSuffix ` field. For example, for the "other eye" transformations (below), the description suffix would need to be `eye_other_spec` for the `dx` fields and `other_spec` for the `cg` fields.
```
  case object OtherEye
      extends HealthCondition(
        198L,
        "Other eye condition",
        Eye,
        cg = Some("other"),
        dx = Some("eye_other")
      )
```
It would be great if we could just pass in the `spec` part, but the "valve" field prevents that (the normal fields are `hs_dx_cardiac_vd_*` and the spec field is `hs_dx_cardiac_valve `.

Let me know what you think, or if it would be easier to do a quick call to talk about it.

danxmoran: Cursed HLES schema... If the way you have it now is working fine, I say we just leave it.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/HealthTransformationsSpec.scala
```diff
@@ -815,6 +815,124 @@ class HealthTransformationsSpec extends AnyFlatSpec with Matchers {
...
+        hsDiagnosisYear = Some(2019),
+        hsDiagnosisMonth = Some(1),
+        hsRequiredSurgeryOrHospitalization = Some(1),
+        hsFollowUpOngoing = Some(true)
+      ),
+      HlesHealthCondition(
+        dogId = 1L,
+        hsConditionType = HealthConditionType.Cardiac.value,
+        hsCondition = HealthCondition.ValveDisease.value,
+        hsConditionOtherDescription = Some("pesto"), // currently None
```
danxmoran, line 909: Is this //currently None comment still accurate?

snwessel: Oh, oops. No, that's out of date, I'll get rid of the comment.

hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/HealthTransformations.scala
quazi-broad: With just two instances it feels acceptable to have the duplicate code

PR #51, opened 2020-07-02
Sw dspdc 878 immune conditions

snwessel:
Adding immune conditions and tests, also updating the names of some skin conditions to avoid naming conflicts.


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/healthcondition/HealthCondition.scala
```diff
@@ -84,20 +84,20 @@ object HealthCondition extends LongEnum[HealthCondition] {
   case object ChronicHotSpots extends HealthCondition(406L, "Chronic or recurrent hot spots", Skin, dx = Some("chs"))
   case object ChronicSkinInfections extends HealthCondition(407L, "Chronic or recurrent skin infections", Skin, dx = Some("csi"))
   case object ContactDermatitis extends HealthCondition(408L, "Contact dermatitis", Skin, dx = Some("cd"))
-  case object DLE extends HealthCondition(409L, "Discoid lupus erythematosus (DLE)", Skin, dx = Some("dle"))
+  case object SkinDLE extends HealthCondition(409L, "Discoid lupus erythematosus (DLE)", Skin, dx = Some("dle"))
```
quazi-broad, line 87: Ah that's a pain. I'm wondering if there's anything I can do to see these types of conflicts up front as I do the oral conditions besides just going through them all and then refactoring as needed. Have we had to do a lot of naming conflicts in other condition types?

quazi-broad: I was also wondering if there was a convention for resolving the conflicts - would it just be adding the condition type as a prefix to the field like you did here?

snwessel: For noticing the conflicts: I started doing a CTRL+F search on the description to make sure it's not used elsewhere (it looks like the description stays consistent throughout). I don't know if that's necessary, but it makes me feel reassured at least. 

And in terms of resolving the conflicts, I believe Raaid did the same renaming/ prefix-changing thing when he ran into this issue on another ticket, so I would say that's the convention for now.

hle-survey/transformation/src/test/scala/org/broadinstitute/monster/dap/HealthTransformationsSpec.scala
```diff
@@ -1381,6 +1381,56 @@ class HealthTransformationsSpec extends AnyFlatSpec with Matchers {
...
+        hsDiagnosisYear = Some(2020),
+        hsDiagnosisMonth = Some(2),
+        hsRequiredSurgeryOrHospitalization = Some(3),
+        hsFollowUpOngoing = Some(true)
+      ),
+      HlesHealthCondition(
+        dogId = 1L,
+        hsConditionType = HealthConditionType.Immune.value,
+        hsCondition = HealthCondition.OtherImmune.value,
+        hsConditionOtherDescription = Some("sriracha"),
```
raaidbroad, line 1509: hahaha

PR #55, opened 2020-07-07
QH-DSPDC-865-health oral

quazi-broad:
Added oral conditions to HealthCondition.
Added unit tests for congential oral disorders and oral diseases.


hle-survey/transformation/src/main/scala/org/broadinstitute/monster/dap/healthcondition/HealthCondition.scala
```diff
@@ -73,7 +73,20 @@ object HealthCondition extends LongEnum[HealthCondition] {
   case object OtherEar extends HealthCondition(298L, "Other ear condition", Ear, both = Some("other"), isOther = true)
 
   // Oral conditions.
-  // TODO
+  case object CleftLip extends HealthCondition(302L, "Cleft lip", Oral, cg = Some("cl_lip"))
```
snwessel, line 76:
```suggestion
  case object CleftLip extends HealthCondition(301L, "Cleft lip", Oral, cg = Some("cl_lip"))
```
Should this start at 301 instead of 302?

quazi-broad: Ah yes, it should, thank you! Good catch!

PR #63, opened 2020-08-10
Extend tsv output script for manual DAP updates

quazi-broad:
Extended the existing python script to break extra wide tables (> 512 cols) into multiple splits. Only applies to hles_dog at the moment but this is a holdover script until we have the ability to automate updating the DAP Terra workspace and won't need the tsv outputs anymore. Left lots of comments, it's pretty messy and probably could be refactored a little but since this is a temporary solution anyways, not sure if it's worth the time. I thought about rewriting the args collection with argparse and going through and making the loops a little easier to read but again, this is just a hacky script anyways. Let me know your thoughts!

Arguments:
-input directory of transform outputs
-output directory for tsvs
-optional string to provide more verbose printing

Sample Calls:
`python convert-output-to-tsv.py /Users/qhoque/DATA/DAP/transform /Users/qhoque/DATA/DAP/tsv`
`python convert-output-to-tsv.py /Users/qhoque/DATA/DAP/transform /Users/qhoque/DATA/DAP/tsv -d`

Other notes:
This PR includes the switch to the new API url. Should I pull that out and put that into it's own PR under a second ticket because that is a completely unrelated task?

Part of this ticket was also to create an ID for the health condition table using existing values in the data. Dan had me use dog_id + health condition but I hit an issue when I tried to upload these to the Terra workspace because there were duplicate values for some dogs (some dogs had two health conditions with the same health condition ID with the difference being whether it was congenital or not. I updated the code to check this flag and include it as an additional int in the derived ID.

### DIFF
--- a/hack/convert-output-to-tsv.py
+++ b/hack/convert-output-to-tsv.py
@@ -24,6 +24,7 @@ def printd(x):
         print("[DEBUG] "+x)
     else:
         pass
+
 # Process the known (hardcoded) tables
 for table_name in table_names:
     print ("\nPROCESSING {0}".format(table_name))
@@ -87,6 +88,7 @@ for table_name in table_names:
     column_set.remove(entity_name)
     sorted_column_set = sorted(list(column_set))    
     sorted_column_set.insert(0, entity_name)
+
     # provide some stats
     col_count = len(sorted_column_set)
     print("...%s contains %s rows and %s columns" % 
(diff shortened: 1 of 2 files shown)


hack/convert-output-to-tsv.py
```diff
@@ -2,38 +2,163 @@
...
-    with open(output_location, 'w') as output_file:
-        dw = csv.DictWriter(output_file, sorted(list(column_set)), delimiter='\t')
-        dw.writeheader()
-        dw.writerows(row_list)
+    # 512 column max limit per request (upload to workspace) 
+    if (col_count > 512):
+        # calculate chunks needed - each table requires the PK
+        total_col_count = ceil(col_count/512)+col_count-1
+        chunks = ceil(total_col_count/512)
+        print("...Splitting %s into %s files" % (table_name, chunks))
```
raaidbroad, line 102: can use the print(f"...Splitting {table_name} into {chunks} files") syntax, could be a little shorter but this is fine too. Your call.

PR #64, opened 2020-09-01
Added a new table under jade-tables to store HLES label mappings

quazi-broad:
Added 3 columns:

column_name (stringified name of the column in the hles_* table)
raw_value (raw number value)
label (string label)

I wasn't sure if the raw_value should be the primary key here or if it should just be the column name.


quazi-broad:
Failing test was due to this branch missing upstream changes (new redcap URL). Changes merged and tests passing successfully now, merging to master.

PR #66, opened 2020-10-15
Adding error handling capability (implemented specifically for missing owner_id)

quazi-broad:
Bumping ingest-sbt-plugins to 2.1.3
Added a PostProcess object to the transformation project
Added a new TransformationLog file to catch Transformation Errors
Added a case class for MissingOwnerId, we can add additional validations as needed
Updated DogTransformations to check the st_owner_id field and proceed based on it's presence or absence
Had to update the Pipeline code to add the PostProcess object
Had to add an implicit logger class to the PipelineBuilder for it to be associated and picked up within DogTransformations
**Still need to add a good unit test for the case of missing st_owner_id**

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
@@ -11,7 +11,7 @@ object DogTransformations {
 
     rawRecord.getOptional("st_owner_id") match {
       case None =>
-        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id")
+        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id").log()
         None
       case Some(ownerId) =>
         Some(
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
@@ -11,13 +11,13 @@ object DogTransformations {
 
     rawRecord.getOptional("st_owner_id") match {
       case None =>
-        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id").log()
+        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id")
         None
-      case Some(ownerId) =>
+      case Some(owner_id) =>
         Some(
           HlesDog(
             dogId = dog_id,
-            ownerId = ownerId.toLong,
+            ownerId = owner_id.toLong,
             hlesDogStudyStatus = Some(StudyStatusTransformations.mapStudyStatus(rawRecord)),
             hlesDogDemographics = Some(DemographicsTransformations.mapDemographics(rawRecord)),
             hlesDogResidences = Some(DogResidenceTransformations.mapDogResidences(rawRecord)),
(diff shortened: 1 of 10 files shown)


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
```diff
@@ -6,24 +6,35 @@ import org.broadinstitute.monster.dogaging.jadeschema.table.HlesDog
...
-        Some(MedsAndPreventativesTransformations.mapMedsPreventatives(rawRecord)),
-      hlesDogHealthSummary = Some(HealthStatusTransformations.mapHealthSummary(rawRecord)),
-      hlesDogFutureStudies = Some(AdditionalStudiesTransformations.mapFutureStudies(rawRecord))
-    )
+  def mapDog(rawRecord: RawRecord): Option[HlesDog] = {
+    val dog_id = rawRecord.id
+
+    rawRecord.getOptional("st_owner_id") match {
+      case None =>
+        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id")
```
pshapiro4broad, line 14: Should this be
```suggestion
        MissingOwnerId(s"Record $dog_id has more/less than 1 value for field st_owner_id").log
```

I don't see how `log` is called otherwise.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/DogTransformationsSpec.scala
```diff
@@ -11,20 +11,22 @@ class DogTransformationsSpec extends AnyFlatSpec with Matchers with OptionValues
 
   it should "map required fields" in {
```
pshapiro4broad, line 12: Would be nice to have a test case that 1) verifies `None` is returned and 2) verifies the counter is incremented. Verifying logging is nice too; not sure how hard it would be to do though.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
pshapiro4broad: It's odd, it looks like you added `log` and then removed it. Was it not needed?

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/OwnerTransformationsSpec.scala
```diff
@@ -1,6 +1,7 @@
 package org.broadinstitute.monster.dap
 
 import org.broadinstitute.monster.dogaging.jadeschema.table.HlesOwner
+import org.scalactic.Fail
```
raaidbroad, line 4: Pretty sure this should be imported from org.scalatest

raaidbroad (commented):
Did you run this on dataflow and observe the desired behavior, i.e. if errors are present then they are logged and the whole pipeline runs but ultimately "fails", and if no errors are present then it runs normally?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/DogTransformations.scala
raaidbroad: @quazi-broad you need to import `import org.broadinstitute.monster.dap.HLESurveyTransformationPipelineBuilder.logger` from the main builder so you have the implicit logger, and then call `.log` on `MissingOwnerId`. I'd also advise renaming `MissingOwnerId` to more explicitly be an error, but that's more my preference/opinion.

quazi-broad: Thank you Phil and Raaid, with the logger from the builder explicitly imported, I am able to call .log on the error now and don't get compile errors (that's why I had added and then removed it prior).

PR #67, opened 2020-11-17
Adding DAP Pack Count and Date as required fields in the initial extraction filter

quazi-broad:
Need to add some sort of validation for this - would it make more sense to test for this in the integration test or unit test (or both?) ?


aherbst-broad:
@quazi-broad both would be good, a unit test would be a good place to start.

PR #68, opened 2020-11-17
Add CSLB schema and update pipelines to reflect new model

aherbst-broad:
## Why
([relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1352))

We want to start pulling in CSLB data from RedCap.

## This PR
* Adds a `CslbExtractionPipeline` class that performs the needed extraction + transform of CSLB data
* Parameterizes the RedCap `event` API parameter so we can pass in a different survey [arm](https://help.redcap.ualberta.ca/help-and-features/what-are-arms). I observed in test runs that we weren't pulling in the needed cslb data. I believe we need to pointing at the `annual_2020_arm_1` arm rather than `baseline_arm_1` for CSLB (note the event name capture on the [CSLB completed report](https://redcap.dogagingproject.org/redcap_v10.5.1/DataExport/index.php?pid=895&report_id=9895) in redcap)
  * Note this will need updating in 2021, we can probably include those arms as well, but I want to get the base ingest working first.


schema/src/main/jade-tables/hles_cslb.table.json
```diff
@@ -0,0 +1,75 @@
+{
+  "name": "hles_cslb",
+  "columns": [
+    {
+      "name": "dog_id",
+      "datatype": "integer",
+      "type": "primary_key"
+    },
+    {
+      "name": "cslb_date",
```
aherbst-broad, line 10: Need to decide on formal terra names for these

aherbst-broad: After discussing with the DAP folks, these should be ok.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/Args.scala
```diff
@@ -10,7 +10,7 @@ import org.broadinstitute.monster.buildinfo.DogAgingHlesExtractionBuildInfo
 
 @AppName("DAP HLE extraction pipeline")
 @AppVersion(DogAgingHlesExtractionBuildInfo.version)
-@ProgName("org.broadinstitute.monster.dap.HLESurveyExtractionPipeline")
+@ProgName("dog-aging-ingest")
```
aherbst-broad, line 13: ❓ I am unable to find any real documentation behind this annotation other than [this example](https://github.com/alexarchambault/case-app#customizing-items-of-the-help--usage-message) in the caseapp readme. It seems like it should only affect the CLI args output, so I am making this change to be more reflective of the actual program name.

quazi-broad (approved):
Changes look good to me, having arm parameterized will be very helpful as we add new instruments - I imagine they'll have different arms, too. Unit tests also look good!
Approving

raaidbroad (approved):
Looks good; given that we're taking the approach of putting different extractions into the extraction directory, instead of taking the approach of having an extraction/transformation/etc per "project" (like HLES vs other things), do we want to restructure the directory structure through the build.sbt file? If I am misunderstanding this and the CLSB stuff is a _part_ of the HLES stuff, then nvm!

aherbst-broad:
@raaidbroad thanks. Yeah that's a great point;  I think I am in favor of an ETL per "project" rather than dumping everything here https://github.com/DataBiosphere/dog-aging-ingest/tree/arh-dspdc-1352-cslb/hles. I'll ticket that up as well as I think we'll need to introduce base abstractions for the transformation piece.

PR #71, opened 2020-11-20
DAP Environmental Data Pull - All Instruments

quazi-broad:
Added 5 jade-fragments, one for each available Redcap instrument under the Environmental Data.

For each fragment, I also created a transformation script to parse those fields and inject them into an Environment record.
I added the 5 additional env data forms to the list of extracted forms in the extraction pipeline builder.
I added code within HLESurveyTransformationPipelineBuilder to call the transformation code.
NOTE with Drew's changes - restructuring and making a generic "ExtractionPipeline" means I would have to slightly rework this code.
The changes required would be to move the EnvironmentalTransformations call outside of the HLES scope (DAP also confirmed this is separate - calculated from address data, not from HLES instrument).


quazi-broad:
Currently IT test on my computerwith the same error we've seen before:
`<b>Fatal error</b>:  Out of memory (allocated 643825664) (tried to allocate 8192 bytes) in <b>D:\home\site\wwwroot\redcap_v10.5.1\Classes\Records.php</b> on line <b>4465</b><br />
{"error":"An unknown error occurred. Please check your API parameters."}
  | => dat org.broadinstitute.monster.common.msg.JsonParser$.$anonfun$parseEncodedJson$1(JsonParser.scala:18)
        at scala.util.Either.fold(Either.scala:192)
        at org.broadinstitute.monster.common.msg.JsonParser$.parseEncodedJson(JsonParser.scala:19)
        at org.broadinstitute.monster.dap.RedCapClient$$anon$1.onResponse(RedCapClient.scala:128)
        at okhttp3.internal.connection.RealCall$AsyncCall.run(RealCall.kt:504)
        at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149)
        ... 2 more
Caused by: io.circe.ParsingFailure: expected json value got '<br />...' (line 1, column 1)
^C
[warn] Canceling execution...`
I believe this always happens when we run the DAP pipeline on our local machine instead of on a DF runner.
Need to confirm that, but if that's true, it should succeed on the automated PR tests.
If it still fails there, why? Automated CI test on a PR creation should be running against dataflow, right?
What would we need to do to make that CI test (more) useful?

quazi-broad:
Interesting, the automated PR run fails with this error:
`Error: Unable to process command '::set-env name=CI,::true' successfully.
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/`
@aherbst-broad could this be relating to any changes we made recently? I wouldn't think so since you haven't pushed anything to master yet, right?

aherbst-broad:
@quazi-broad, no nothing pushed to master. However, check out my [upgrade](https://github.com/DataBiosphere/dog-aging-ingest/blob/9de502047bf5d28f1617d57b9280d10bac939b32/.github/workflows/validate-pull-request.yaml#L11) of the setup-scala github action in my PR, that fixes your second issue.

Re: the IT failures: I run into those ~50% of the time locally and in CI. I re-ran until they passed and got the ✅ . Very not great, but I am uncomfortable ignoring these tests. I _believe_ the CI tests are just running in memory via this [github action](https://github.com/DataBiosphere/dog-aging-ingest/blob/master/.github/workflows/validate-pull-request.yaml#L23) (the same code path as when we run locally).

I think we should have a quick huddle this week and brainstorm out a plan on making these tests not block us so consistently.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/TemperaturePrecipitationTransformations.scala
```diff
@@ -0,0 +1,116 @@
...
+  /**
+    * Parse all temperature and precipitation variables out of a raw RedCap record,
+    * injecting them into a partially-modeled environment record.
+    */
+  def mapTemperaturePrecipitationVariables(
+    rawRecord: RawRecord
+  ): EnvironmentTemperaturePrecipitation = {
+    EnvironmentTemperaturePrecipitation(
+      tpDataYear = rawRecord.getOptionalNumber("tp_data_year"),
+      tpPcpnAnnual01 = rawRecord.getOptionalNumber("tp_pcpn_annual_01").map(_.toDouble),
```
raaidbroad, line 17: it would be so great if we didn't have to write all this out since it is so repeateddd ugh :/ I don't think we have a good way to do so thus far

raaidbroad (approved):

aherbst-broad (changes requested):
Looks good! I think this needs unit tests for the transformations.

quazi-broad:
This PR has gotten too large and is not so digestible.
These code changes were split up into 3 separate PRs:
Schema Changes:
https://github.com/DataBiosphere/dog-aging-ingest/pull/80
Extraction Code:
https://github.com/DataBiosphere/dog-aging-ingest/pull/82
Transformation Code:
TBD

PR #76, opened 2020-12-01
Shift to new ingest-utils json parsing method

aherbst-broad:
## Why

The DAP pipeline does not terminate cleanly if an exception is thrown from the Redcap client (usually if non-JSON garbage is returned). We should handle that situation and terminate.

## This PR
* Shifts to a new json parsing method introduced in ingest-utils 2.1.4 that returns an `Either[ParsingFailure, Msg]` that we will match against.

### DIFF
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
@@ -104,12 +104,12 @@ class ExtractionPipelineBuilder(
         _.applyKvTransform(ParDo.of(lookupFn)).flatMap(kv => kv.getValue.fold(throw _, _.arr))
       }
 
-    StorageIO.writeJsonLists(
+    StorageIO.writeJsonListsGeneric(
       extractedRecords,
       "Write records",
       s"${args.outputPrefix}/${subDir}/records"
     )
-    StorageIO.writeJsonLists(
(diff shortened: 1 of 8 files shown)


raaidbroad (approved):

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
```diff
@@ -104,12 +104,12 @@ class ExtractionPipelineBuilder(
...
       }
 
-    StorageIO.writeJsonLists(
+    StorageIO.writeJsonListsGeneric(
       extractedRecords,
       "Write records",
       s"${args.outputPrefix}/${subDir}/records"
     )
-    StorageIO.writeJsonLists(
+    StorageIO.writeJsonListsGeneric(
```
ThePletch, line 112: What's the distinction here?

ThePletch (approved):

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
aherbst-broad: This is a new method @raaidbroad introduced in the latest version of ingest-utils. Briefly, writeJsonLists is generic in `[M : Encoder]` while `writeJsonListsGeneric` is typed to `SCollection[Msg]`. I think this is effectively the same codepath, just a new name.

PR #77, opened 2020-12-01
[DSPDC-1331] Implement DAP pack filtering

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1331)

We want to filter for 
* the presence of `st_dap_pack_count`
* the presence of `st_dap_pack_date`

In my testing, the REDCap API does not support "not null" conditions, so I've implemented something more hacky where we check for `st_dap_pack_count` > 0 and `st_dap_pack_date` > some magic date in the past.

## This PR
* We had hardcoded our filtering logic in the redcap client to only support `=`. I've extracted the filtering directives to their own typesafe class so we can now support other operands (">" for now).
* Updates the CSLB + HLES pipelines with the new filtering approach, and adds the DAP pack filters to the HLES pipeline.

### DIFF
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/CslbExtractionPipeline.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/CslbExtractionPipeline.scala
@@ -13,8 +13,8 @@ object CslbExtractionPipeline extends ScioApp[Args] {
   // Magic marker for "completed".
   // NB: We are purposefully excluding the recruitment_fields_complete -> 2
   // mapping, as that conflicts with the CSLB data
-  val extractionFilters: Map[String, String] = Map(
-    s"canine_social_and_learned_behavior_complete" -> "2"
+  val extractionFilters: List[FilterDirective] = List(
+    FilterDirective("canine_social_and_learned_behavior_complete", FilterOps.==, "2")
   )
 
   val subdir = "cslb"
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipeline.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipeline.scala
@@ -9,6 +9,8 @@ import Args._
 object HLESurveyExtractionPipeline extends ScioApp[Args] {
 
   /** Names of all forms we want to extract as part of HLE ingest. */
+  val HLESEpoch = "2018-01-01"
+
   val forms = List(
     "recruitment_fields",
     "owner_contact",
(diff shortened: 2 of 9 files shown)


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/CslbExtractionPipeline.scala
```diff
@@ -13,8 +13,8 @@ object CslbExtractionPipeline extends ScioApp[Args] {
   // Magic marker for "completed".
   // NB: We are purposefully excluding the recruitment_fields_complete -> 2
   // mapping, as that conflicts with the CSLB data
-  val extractionFilters: Map[String, String] = Map(
-    s"canine_social_and_learned_behavior_complete" -> "2"
+  val extractionFilters: List[FilterDirective] = List(
+    FilterDirective("canine_social_and_learned_behavior_complete", FilterOps.==, "2")
```
aherbst-broad, line 17: @quazi-broad should CSLB reflect the DAP pack checks as well?

quazi-broad: I actually do not think so but we briefly talked about this when we were discussing changing the extraction filters. In our chat you noted "it sure seems like HLES + CSLB extraction criteria are mutually exclusive". I don't think we got around to actually asking Matt that so I'll add that to the list of things to ask him. When we call the CSLB extraction, does it pull all the HLES fields down, or just the CSLB subset. If it's just the CSLB subset, then I do not think we can use the same filters as some of the fields that we check do not exist in the CSLB subset. I'll bring this up with Matt

ThePletch (approved):

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/HLESurveyExtractionPipeline.scala
```diff
@@ -9,6 +9,8 @@ import Args._
 object HLESurveyExtractionPipeline extends ScioApp[Args] {
 
   /** Names of all forms we want to extract as part of HLE ingest. */
+  val HLESEpoch = "2018-01-01"
```
quazi-broad, line 12: Just curious about this date - is this the earliest possible DAP pack date?

aherbst-broad: A very good question, this was me being paranoid and choosing what I thought was an "early enough" date. I just pulled down the DAP Pack report someone had setup in RC and noted that the earliest date is 2019-12-23 so I will change to that.

quazi-h (approved):
These changes look good to me, I like that this will allow us to feed in specific extraction filters and we can specify the exact logic for the filters for each specific pipeline.

PR #78, opened 2020-12-02
Add unit tests for HLES transform pipeline, fix double log for missing owner ID, add test fixture/logging config

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1333)

## This PR
* Adds unit tests for the HLES transformation pipeline's skipping of records missing the `st_owner_id` field
* Fixes the double-reporting of errors for such fields across owner and dog transforms
* Adds configuration for logging in transform tests
* Adds an example of using local test fixtures in pipeline tests


hles/transformation/src/test/resources/org/broadinstitute/monster/dap/blank_st_owner_id/records/two_blank_owner_ids_of_three_records.json
```diff
@@ -0,0 +1,1473 @@
+{"record":"87616","redcap_event_name":"baseline_arm_1","field_name":"study_id","value":"87616"}
```
aherbst-broad, line 1: Any PII should be redacted out of this fixture data if it hasn't already been.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,75 @@
+package org.broadinstitute.monster.dap
+
+//import better.files._
```
aherbst-broad, line 3: 🪓 this.

hles/transformation/src/test/resources/org/broadinstitute/monster/dap/blank_st_owner_id/records/two_blank_owner_ids_of_three_records.json
ThePletch: good point, i'll give it a pass

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
```diff
@@ -35,7 +33,7 @@ object OwnerTransformations {
 
     rawRecord.getOptional("st_owner_id") match {
       case None =>
-        MissingOwnerIdError(s"Record $dog_id has less than 1 value for field st_owner_id").log
+        // we log an error for missing st_owner_id during dog transformation, so we just skip the record here.
```
quazi-broad, line 36: looks good, good catch!

hles/transformation/src/test/resources/logback.xml
```diff
@@ -0,0 +1,32 @@
+<configuration>
```
aherbst-broad, line 1: ❓ Should this be in `logback-test.xml`?

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+  ): scalatest.Assertion = {
+    val result = runWithRealContext(opts)(ctx =>
+      HLESurveyTransformationPipelineBuilder.buildPipeline(
+        ctx,
+        Args(
+          this.getClass.getResource(inputPath).toString,
+          outputDir.pathAsString
+        )
+      )
+    ).waitUntilFinish()
```
aherbst-broad, line 45: ❓ I've been trying to get my head around this with `ScioContext` as well: why `waitUntilFinish` vs `waitUntilDone`?

ThePletch: This is replicated from the HCA pipeline validation tests - @raaidbroad can you help us understand the distinction here?

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    val src = Source.fromFile("../logs/errors.log")
+    val lines = src.getLines.toList
+    src.close()
+    lines.map(err => JsonParser.parseEncodedJson(err)).takeRight(count) should contain(
+      expectedErrorMessageJson
+    )
+
+    // make sure counter was incremented right number of times
+    val errorCountMetric: MetricName = MetricName.named("main", errCount)
+    assert(result.allCounters.isDefinedAt(errorCountMetric))
```
aherbst-broad, line 58: ⛏️ it'd be good to use a matcher for consistency with the rest of our test suite, rather than an assertion. i.e., `result.allCounters should contain (errorCountMetric)`

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+    result.allCounters.foreach {
```
aherbst-broad, line 59: ⛏️ I'm not sure a full scan over the map is needed here; if we already have the metric name we're after (`errorCountMetric`), can't we just pull the value and make the assertion? i.e,. 
`result.allCounters(errorCountMetric).attempted shouldBe count`

ThePletch: We can, but as far as I can tell, pulling the value returns an `Option[MetricValue]`, which i then have to scan over anyway. I'm still learning what's idiomatic in Scala.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+      case (name, errorCount) =>
+        if (name == errorCountMetric) errorCount.attempted shouldBe count
+    }
+
+    assertThrows[HLESurveyTransformationFailException](PostProcess.postProcess(result))
```
aherbst-broad, line 64: ⛏️ I think you can use a matcher here as well https://www.scalatest.org/user_guide/using_matchers#expectedExceptions

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
aherbst-broad: I think these checks should be part of additional tests (i.e,. `it should "increment error counters properly"` instead of intermixing them with the pipeline setup code.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
aherbst-broad: Fair; it's an option in a test, you can just do a `get()` on the value and make the assertion then.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+  // set up a pipeline output directory that we can then clean up
+  val outputDir = File.newTemporaryDirectory()
+  override def afterAll(): Unit = outputDir.delete()
+
+  def pipelineTest(
+    inputPath: String,
+    expectedErrorMessage: String,
+    count: Int = 1
+  ): scalatest.Assertion = {
+    val result = runWithRealContext(opts)(ctx =>
```
aherbst-broad, line 36: 🌱 I like this pattern (i.e., feeding fixture data from the test resources dir into the pipeline, rather than making live calls against the API). I think we should convert our ITs over to this once approach once this is merged.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
raaidbroad: `waitUntilDone` Does the following: Wait until the pipeline finishes with the State `DONE` (as opposed to `CANCELLED` or `FAILED`). Throw exception otherwise.

`waitUntilFinish` doesn't throw an exception if the pipeline finishes with a different state. I think using `waitUntilFinish` is a relic of when the approach was to actually throw an exception and break the pipeline; since the approach we take in the hca pipeline is now to log, drop, and carry-on, we don't need to use `waitUntilFinish` to ensure the tests run, sooo might as well use `waitUntilDone` now. Good question.

raaidbroad: There is a non-zero chance that this also has something to do with running the tests on Dataflow; I ran into some weirdness with tests running locally with the direct runner but then breaking with the dataflow runner. I'ma go check the PR history

raaidbroad: Don't see any comments/commits that reference dataflow weirdness for that bit. Feel free to test the difference at your own peril 😇

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
ThePletch: Was not aware of the existence of `get` - I'll use that here instead. I was looking for a way to avoid doing this mapping anyway.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
ThePletch: Bafflingly, changing it to `waitUntilDone` appears to have broken two unrelated specs: `org.broadinstitute.monster.dap.CslbExtractionPipelineBuilderIntegrationSpec` and `org.broadinstitute.monster.dap.HLESExtractionPipelineBuilderIntegrationSpec`, which failed with this opaque stack trace: ```
[info] org.broadinstitute.monster.dap.HLESExtractionPipelineBuilderIntegrationSpec *** ABORTED ***
[info]   org.apache.beam.sdk.Pipeline$PipelineExecutionException: io.circe.ParsingFailure: expected json value got '<br />...' (line 1, column 1)
[info]   at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:348)
[info]   at org.apache.beam.runners.direct.DirectRunner$DirectPipelineResult.waitUntilFinish(DirectRunner.java:318)
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:213)
[info]   at org.apache.beam.runners.direct.DirectRunner.run(DirectRunner.java:67)
[info]   at org.apache.beam.sdk.Pipeline.run(Pipeline.java:317)
[info]   at org.apache.beam.sdk.Pipeline.run(Pipeline.java:303)
[info]   at com.spotify.scio.ScioContext.execute(ScioContext.scala:609)
[info]   at com.spotify.scio.ScioContext$$anonfun$run$1.apply(ScioContext.scala:597)
[info]   at com.spotify.scio.ScioContext$$anonfun$run$1.apply(ScioContext.scala:585)
[info]   at com.spotify.scio.ScioContext.requireNotClosed(ScioContext.scala:674)
[info]   ...
[info]   Cause: io.circe.ParsingFailure: expected json value got '<br />...' (line 1, column 1)
[info]   ...
```

aherbst-broad: @ThePletch that error is very likely unrelated and due to a 500 on the Redcap server (this happens to me a bunch too, and we need to fix it)

aherbst-broad (approved):
lgtm

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineValidationSpec.scala
quazi-broad: Yup I think Drew correctly diagnosed that - I've hit that issue in the past before without making any changes to the code.

quazi-h (approved):
Looks good to me!

raaidbroad (commented):
🤔 Do we want to add GH action steps to run the tests on dataflow on a merge to master? Kind of like here: https://github.com/DataBiosphere/hca-ingest/blob/9cc9a062b9f601d8959d3cd9c8bc145e45a6e0d2/.github/workflows/build-and-publish-master.yaml#L27-L28 or would we want that to be a separate PR?

PR #82, opened 2020-12-09
Add extraction code to pull environment data

quazi-h:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/1287)
This is the second PR in a series to get the DAP environmental data extracted from RedCap.
We are breaking a larger more convoluted PR into smaller pieces and this is the second of that series.

## This PR

-Parameterized "arm" so that it can be passed as an arguent per pipeline ("events" in RedCapClient)
-Parameterized "fieldList" so that it can be passed as an arguent per pipeline ("fields" in RedCapRequest)
-Updated HLES, CSLB and PipelineBuilder unit test to pass newly added arm and fieldList parameters
-Added a new ExtractionPipeline script for Environment

### DIFF
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
@@ -1,0 +1,40 @@
+package org.broadinstitute.monster.dap
+
+import org.broadinstitute.monster.common.{PipelineBuilder, ScioApp}
+
+// Ignore IntelliJ, this is used to make the implicit parser compile.
+import Args._
+
+object EnvironmentExtractionPipeline extends ScioApp[Args] {
+
+  val forms = List(
+    "geocoding_metadata",
+    "census_variables",
+    "pollutant_variables",
+    "temperature_and_precipitation_variables",
+    "walkability_variables"
+  )
+
+  // Magic marker for "completed".
+  // NB: We are looking for baseline_complete -> 2
+  val extractionFilters: List[FilterDirective] = List(
+    FilterDirective("baseline_complete", FilterOps.==, "2")
+  )
+
+  val subdir = "environment"
+  //todo: need to query for all arms and work through arms serially
+  val arm =
+    List("baseline_arm_1", "dec2019_arm_1", "jan2020_arm_1")
+  val fieldList = List("baseline_complete")
+
+  override def pipelineBuilder: PipelineBuilder[Args] =
+    new ExtractionPipelineBuilder(
+      forms,
+      extractionFilters,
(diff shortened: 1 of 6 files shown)


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
```diff
@@ -87,7 +89,7 @@ class ExtractionPipelineBuilder(
           ids = ids.getValue.asScala.toList,
           forms = formsForExtraction,
           // Pull the consent field so we can QC that the filter is working properly.
```
aherbst-broad, line 91: ⛏️ This comment doesn't make a lot of sense given this change.

aherbst-broad (approved):
Looks good, nice job breaking this into smaller prs.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -0,0 +1,40 @@
+package org.broadinstitute.monster.dap
+
+import org.broadinstitute.monster.common.{PipelineBuilder, ScioApp}
+
+// Ignore IntelliJ, this is used to make the implicit parser compile.
+import Args._
+
+object EnvironmentExtractionPipeline extends ScioApp[Args] {
```
ThePletch, line 8: where does this pipeline get invoked?

ThePletch (commented):
pipeline itself looks good, but i'm a bit unclear on how it actually gets called

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
quazi-broad: When we call extraction on sbt, we are presented with 3 different choices for which pipeline to run (HLES, Environment, CSLB) and the user is prompted to choose one of the three options. When the user chooses one of the pipelines, I think this code is then executed - would that be the point where the specific pipeline gets invoked? I see that ScioApp has the following function:
`
  /**
    * Entry-point for scio CLPs.
    *
    * Parses command-line args, constructs a pipeline, runs the
    * pipeline, and waits for the pipeline to finish.
    */
  def main(args: Array[String]): Unit = {
    val (pipelineContext, parsedArgs) =
      ContextAndArgs.typed[Args](args)(parser.nameFormatter(argFormatter), help)
    pipelineBuilder.buildPipeline(pipelineContext, parsedArgs)
    val result = pipelineContext.run().waitUntilDone()
    postProcess(result)
  }`

ThePletch (approved):

PR #85, opened 2020-12-16
Add DAP ENV transformation code

quazi-h:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/1287)
[Bug ticket](https://broadinstitute.atlassian.net/browse/1370)

This is the third and final PR in a series to get the DAP environmental data extracted from RedCap.
We are breaking a larger more convoluted PR into smaller pieces and this is the last of that series.
We are also addressing a duplicate records issue in this PR.

## This PR
-Added 5 transformation scripts for the env data subsets
-Added a unit test for each type of env transformation
-Refactored HLESurveyTransformationPipelineBuilder to process environmental records separately
-Updated the hardcoded list of arms to include largest subset we can grab successfully
-Addressed duplicates (DSPDC-1370) in data by adding some logic to RawRecord.
-Added logic to process the Environment table in the tsv script under /hack
-Expanded the list of arms to environment to include the full set of events in RedCap we should be pulling.
-Modified the client code to take "arm" as a param
-Added a filter for "NA" values when getOptionalDateTime is used
-Added logic to remove "NA" values from walkscoreDates
-Added unit test cases for new logic and fixed some unit test bugs
-Adjusted the idBatchSize for Environment to 10 to accommodate RedCap
-Split up the pull for initRequests to pull a list of records per arm


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -77,4 +86,34 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+      )
+
+    // Group by study ID (record number) and arm (address_year_month)
+    // to get the format: (studyId, arm_id, Iterable((fieldName, Iterable(value))))
+    // (study_id, arm_id, iterable())
+    rawRecords
+      .groupBy(foo => {
+        (foo.read[String]("record"), foo.read[String]("redcap_event_name"))
+      })
+      //.groupBy(read[String]("redcap_event_name"))
```
aherbst-broad, line 106: Remove this commented out code

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -77,4 +86,34 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+        ctx,
+        "Raw Records",
+        s"${args.inputPrefix}/records/*.json"
+      )
+
+    // Group by study ID (record number) and arm (address_year_month)
+    // to get the format: (studyId, arm_id, Iterable((fieldName, Iterable(value))))
+    // (study_id, arm_id, iterable())
+    rawRecords
+      .groupBy(foo => {
```
aherbst-broad, line 103: 🔧 Can you give this a more descriptive name than `foo`? Perhaps `record`?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -77,4 +86,34 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+      //.groupBy(read[String]("redcap_event_name"))
+      .map {
+        case ((id, eventName), rawRecordValues) =>
+          val fields: Map[String, Array[String]] = rawRecordValues
+            .groupBy(_.read[String]("field_name"))
+            .map {
+              case (fieldName, rawValues) =>
+                (fieldName, rawValues.map(_.read[String]("value")).toArray.sorted)
+            }
+          val bar = fields + ("redcap_event_name" -> Array(eventName))
```
aherbst-broad, line 115: 🔧 Same comment re: naming; you might do away with this intermediate val altogether and do the concat inline when you construct the `RawRecord` instance.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -22,7 +22,7 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
     */
   def getRequired(field: String): String = {
     val values = fields.getOrElse(field, Array.empty)
-    if (values.length != 1) {
+    if (values.toSet.size != 1) {
```
aherbst-broad, line 25: ❤️

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformations.scala
```diff
@@ -0,0 +1,34 @@
...
+    * injecting them into a partially-modeled environment record.
+    */
+  def mapWalkabilityVariables(rawRecord: RawRecord): EnvironmentWalkability = {
+    var walkscore: Option[Double] = None
+    var walkscoreDate: Option[String] = None
+
+    if (!rawRecord.getOptional("wv_walkscore").contains("NA")) {
+      walkscore = rawRecord.getOptional("wv_walkscore").map(_.toDouble)
+    }
+    if (rawRecord.getArray("wv_walkscore_date").nonEmpty) {
```
aherbst-broad, line 19: Can you add a comment describing why this nested conditional exists (I think I remember from our pairing, but we will for sure forget a month from now).

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformations.scala
```diff
@@ -0,0 +1,34 @@
...
+      if (!rawRecord.getArray("wv_walkscore_date")(0).contains("NA")) {
+        walkscoreDate = Some(rawRecord.getArray("wv_walkscore_date")(0))
+      }
+    }
+
+    EnvironmentWalkability(
+      wvWalkscore = walkscore, // rawRecord.getOptional("wv_walkscore").map(_.toDouble),
```
aherbst-broad, line 26: 🔧 Remove commented out code.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformationsSpec.scala
```diff
@@ -0,0 +1,37 @@
+package org.broadinstitute.monster.dap.dog
+
+import org.broadinstitute.monster.dap.RawRecord
+import org.broadinstitute.monster.dap.environment.WalkabilityTransformations
+import org.scalatest.OptionValues
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+
+class WalkabilityTransformationsSpec extends AnyFlatSpec with Matchers with OptionValues {
```
aherbst-broad, line 9: 🔧 I think this test should also exercise the conditional codepaths around `wv_walkscore_date` and `wv_walkscore`

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -5,6 +5,7 @@ import com.spotify.scio.values.SCollection
 import org.broadinstitute.monster.common.{PipelineBuilder, StorageIO}
 import org.broadinstitute.monster.common.msg._
 import org.slf4j.{Logger, LoggerFactory}
+import upack.Msg
 
 object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
```
aherbst-broad, line 10: 🌱 Out of scope for this PR but: we need to do some refactoring after this ships, we're now performing HLES, CSLB and env transforms in a class that is ostensibly for HLES transforms (from it's name at least)

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -77,4 +86,34 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
           RawRecord(id.toLong, fields)
       }
   }
+
+  /** Read in records and group by study Id, with field name subgroups. */
+  def readEnvRecords(ctx: ScioContext, args: Args): SCollection[RawRecord] = {
+    val rawRecords: SCollection[Msg] = StorageIO
+      .readJsonLists(
+        ctx,
+        "Raw Records",
```
aherbst-broad, line 95: ⛏️ "Raw Environmental Records" perhaps?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
aherbst-broad: Can you add a test somewhere that exercises this de-duping behavior?

aherbst-broad (commented):
Looking good so far!

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -24,7 +24,23 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
...
+      "feb2020_arm_1",
+      "feb2020_secondary_arm_1",
+      "mar2020_arm_1",
+      "mar2020_secondary_arm_1",
+      "apr2020_arm_1",
+      "apr2020_secondary_arm_1",
+      "may2020_arm_1",
+      "may2020_secondary_arm_1",
+      "june2020_arm_1",
+      "june2020_secondary_arm_1"
```
aherbst-broad, line 42: ❓ Does this only go to June 2020 for performance reasons?

quazi-broad: yes - adding any more arms makes this fail when running a full refresj

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
quazi-broad: agreed - I really wanted to break out the transformation into 3 streams in this PR but figured we should keep it simple since this is already 1/3 of a larger PR and there's a bunch of stuff being added in this PR alone

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformations.scala
quazi-broad: We observed instances of "NA" in data returned from RedCap - thanks I will add the comment so we don't scratch our heads in the future

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
ThePletch: do we have any mechanisms to throttle this to let us add all of the arms we need without beating redcap into submission? this seems like a high priority feature to add if it's stopping us from doing a proper full refresh.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
```diff
@@ -0,0 +1,27 @@
...
+import org.broadinstitute.monster.dap.environment._
+import org.broadinstitute.monster.dogaging.jadeschema.table.Environment
+
+object EnvironmentTransformations {
+
+  /** Parse all environment related fields out of a RedCap record.
+    * The schema for environment variables has been separated into 5 jade-fragments.
+    */
+  def mapEnvironment(rawRecord: RawRecord): Option[Environment] = {
+    val dog_id = rawRecord.id
```
ThePletch, line 12: :snake: -> :camel:
```suggestion
    val dogId = rawRecord.id
```

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
```diff
@@ -0,0 +1,27 @@
...
+    Some(
+      Environment(
+        dogId = dog_id,
+        addressMonthYear = rawRecord.getRequired("redcap_event_name"),
+        environmentGeocoding = Some(GeocodingTransformations.mapGeocodingMetadata(rawRecord)),
```
ThePletch, line 17: for my understanding, what's the purpose of wrapping these in `Some`?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -77,4 +86,32 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
...
+      .groupBy(record => {
+        (record.read[String]("record"), record.read[String]("redcap_event_name"))
+      })
+      .map {
+        case ((id, eventName), rawRecordValues) =>
+          val fields: Map[String, Array[String]] = rawRecordValues
+            .groupBy(_.read[String]("field_name"))
+            .map {
+              case (fieldName, rawValues) =>
+                (fieldName, rawValues.map(_.read[String]("value")).toArray.sorted)
```
ThePletch, line 112: this block (converting from redcap data structures to something more useful to us) seems like it's something we might want to be able to reuse. can we pull this functionality out into a `RedcapRecordDigester` (working title) object or something similar?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -33,10 +37,13 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
     * Get the singleton value for an attribute in this record, if one exists.
     *
     * Raises an error if the attribute has multiple values.
+    *
+    * Only unique duplicates will throw an error
+    * If we receive multiple fields with the same value, the head of the set is used
```
ThePletch, line 42: is it necessarily safe to discard the extra data if we get multiple unique values for a field?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/CensusTransformations.scala
```diff
@@ -0,0 +1,43 @@
...
+      cvPctFamsownchildFemaleLed =
+        rawRecord.getOptional("cv_pctfamsownchildfemaleled").map(_.toDouble),
+      cvPctLessThanBaDegree = rawRecord.getOptional("cv_pctlessthanba").map(_.toDouble),
+      cvPctLessThan100k = rawRecord.getOptional("cv_pctlessthan100k").map(_.toDouble),
+      cvDisadvantageIndex = rawRecord.getOptional("cv_disadvantageindex").map(_.toDouble),
+      cvPctSameHouse1yrago = rawRecord.getOptional("cv_pctsamehouse1yrago").map(_.toDouble),
+      cvPctOwnerOccupied = rawRecord.getOptional("cv_pctowneroccupied").map(_.toDouble),
+      cvPctUsBorn = rawRecord.getOptional("cv_pctborninus").map(_.toDouble),
+      cvStabilityIndex = rawRecord.getOptional("cv_stabilityindex").map(_.toDouble),
+      cvDataYear = rawRecord.getOptionalNumber("cv_data_year")
```
ThePletch, line 40: definitely not in scope for this PR, but this style of transformation definition feels _really_ verbose. finding myself wondering if we could define transformations more declaratively than this, e.g. just `Map("cv_pctowneroccupied" -> Option[Double], ...)`

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/TemperaturePrecipitationTransformations.scala
```diff
@@ -0,0 +1,116 @@
...
+      tpPcpnAnnual03 = rawRecord.getOptional("tp_pcpn_annual_03").map(_.toDouble),
+      tpPcpnAnnual04 = rawRecord.getOptional("tp_pcpn_annual_04").map(_.toDouble),
+      tpPcpnAnnual05 = rawRecord.getOptional("tp_pcpn_annual_05").map(_.toDouble),
+      tpPcpnAnnual06 = rawRecord.getOptional("tp_pcpn_annual_06").map(_.toDouble),
+      tpPcpnAnnual07 = rawRecord.getOptional("tp_pcpn_annual_07").map(_.toDouble),
+      tpPcpnAnnual08 = rawRecord.getOptional("tp_pcpn_annual_08").map(_.toDouble),
+      tpPcpnAnnual09 = rawRecord.getOptional("tp_pcpn_annual_09").map(_.toDouble),
+      tpPcpnAnnual10 = rawRecord.getOptional("tp_pcpn_annual_10").map(_.toDouble),
+      tpPcpnAnnual11 = rawRecord.getOptional("tp_pcpn_annual_11").map(_.toDouble),
+      tpPcpnAnnual12 = rawRecord.getOptional("tp_pcpn_annual_12").map(_.toDouble),
```
ThePletch, line 28: is it important that we define these individually in sequence, or is there some way we could just pass in all of these values in a loop for each month, 1-12?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformations.scala
ThePletch: could we include handling for `"NA"` values in our various `getOptional/getArray/getWhatever` methods rather than explicitly handling that here, or would that potentially swallow cases where `"NA"` is a valid value for a field?

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/environment/TemperaturePrecipitationTransformationsSpec.scala
```diff
@@ -0,0 +1,223 @@
...
+      "tp_tmin_norm_03" -> Array("33.7"),
+      "tp_tmin_norm_04" -> Array("36.6"),
+      "tp_tmin_norm_05" -> Array("41.7"),
+      "tp_tmin_norm_06" -> Array("46.6"),
+      "tp_tmin_norm_07" -> Array("51"),
+      "tp_tmin_norm_08" -> Array("51.4"),
+      "tp_tmin_norm_09" -> Array("47.1"),
+      "tp_tmin_norm_10" -> Array("40.5"),
+      "tp_tmin_norm_11" -> Array("34.9"),
+      "tp_tmin_norm_12" -> Array("30.3")
```
ThePletch, line 115: at least in the specs, i think we should definitely be building these massive maps programmatically

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/environment/WalkabilityTransformationsSpec.scala
```diff
@@ -0,0 +1,56 @@
...
+    )
+
+    val walkabilityDataMapped = WalkabilityTransformations.mapWalkabilityVariables(
+      RawRecord(1, walkabilityData)
+    )
+
+    // output of the example record's walkability transformations
+    walkabilityDataMapped.wvWalkscore shouldBe Some(76.0)
+    walkabilityDataMapped.wvWalkscoreDescrip.value shouldBe 2L
+    // todo: CHECK THIS - CONFIRM WITH DAP ON TYPE
```
ThePletch, line 31: flagging this todo to be addressed before merging

ThePletch (commented):
couple of minor nits, questions, and down-the-line refactor suggestions

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
aherbst-broad: No, there is no such mechanism now; @quazi-broad and I have been talking about this, we'll tackle it after this gets merged (@quazi-broad can you make sure there's a ticket for this and link it here?)

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/TemperaturePrecipitationTransformations.scala
aherbst-broad: This could get sticky; we codegen (via a plugin from `ingest-utils`) the underlying case class `EnvironmentTemperaturePrecipitation` from a JSON schema. We could muck with the case class JSON encoder to do the array transformation as you allude to, but it'd get clobbered on the next compile.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
quazi-broad: To get the Option[] type, I believe

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
quazi-broad: https://broadinstitute.atlassian.net/browse/DSPDC-1415

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
quazi-broad: I think the way we have this written - we will throw an error from unique duplicates.
This code should only weed out instances with the same value.

aherbst-broad (approved):
lgtm

hack/convert-output-to-tsv.py
```diff
@@ -31,134 +31,149 @@ def printd(x):
...
+                # this thread is executed for any tables with 512 cols or less
+                print(f"...No need to split files for {table_name}")
+                output_location = output_dir + '/' + table_name + '.tsv'
+                print(f"...Writing {table_name} to {output_location}")
+                with open(output_location, 'w') as output_file:
+                    dw = csv.DictWriter(output_file, sorted_column_set, delimiter='\t')
+                    dw.writeheader()
+                    dw.writerows(row_list)
+                print(f"...{table_name} was successfully written to {output_location}")
+    except OSError as e:
```
ThePletch, line 178: :-1: big try-catch blocks like this are a code smell and can make issues harder to debug. let's try to identify lines where we expect errors and want to skip over them (do we, since all we're doing is printing the error?) and put try-catch blocks around them instead.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -35,7 +62,7 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
       arm,
       fieldList,
       subdir,
-      100,
+      10,
```
ThePletch, line 65: let's add a comment about what led us to pick a value of 10 here.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -36,11 +42,20 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
...
-    if (values.length > 1) {
-      throw new IllegalStateException(s"Record $id has multiple values for field $field")
+    val values = fields.getOrElse(field, Array.empty).toSet
+    // If there are multiple values
+    if (values.size > 1) {
+      // Filtering out "NA"
+      if (values.--(Array("NA")).size > 1) {
+        throw new IllegalStateException(s"Record $id has multiple values for field $field")
+      } else {
+        values.headOption
```
ThePletch, line 57: are we worried about this possibly returning `NA`?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -52,11 +67,13 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
...
 
-  /** Get the singleton value for an attribute in this record, parsed as a long. */
+  /** Get the singleton value for an attribute in this record, parsed as a long.
+    * The conversion to double and then long is to handle scientific notation
+    */
   def getOptionalNumber(field: String, truncateDecimals: Boolean = true): Option[Long] =
     getOptional(field).map(value => {
       try {
-        value.toLong
+        value.toDouble.toLong
```
ThePletch, line 76: :-1: if we convert to a double in this try statement, it'll also handle any decimal values, which will prevent us from logging that we needed to truncate a decimal value and instead simply do so silently.

ThePletch (changes requested):
one thing that we need to change to avoid clobbering existing functionality, then a few suggestions for improvements

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
quazi-broad: Moved the toDouble into the TruncatedDecimalError logic

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
quazi-broad: Yes, we were seeing `NA` values being pulled by getOption

hack/convert-output-to-tsv.py
quazi-broad: I've walked back these changes minus just adding the environment table logic.

ThePletch (approved):

PR #86, opened 2020-12-16
Mock out Redcap interactions in extraction pipeline integration tests

ThePletch:
## Why

Our extraction pipelines' integration tests were making live calls to the Redcap API, which meant that our tests would fail if the API went down (which it often did) and that we ended up testing their API's functionality rather than our own code. This change lets us control what is returned from their API by using a mocking layer to specify custom responses for a given request.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1383)

## This PR
* Pulls out HTTP requests into a wrapper, which can either make live requests or be replaced with a mock wrapper during testing
* Adds some utilities to generate both Redcap requests and Redcap responses from simpler data structures
* Removes some of our integration tests that were merely testing the Redcap API's response format rather than our code
* Factors out some parts of the Redcap client that generate raw requests from Redcap request objects to expose them to our test code


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
```diff
@@ -29,115 +26,100 @@ trait RedCapClient extends Serializable {
...
-            val maybeResult =
-              JsonParser.parseEncodedJsonReturningFailure(response.body().string())
-            maybeResult match {
-              case Right(result) => p.success(result)
-              case Left(err)     => p.failure(err)
-            }
-          }
-        })
-      p.future
+  /** Construct a client instance backed by the production RedCap instance. */
```
ThePletch, line 118: i don't know that this comment is accurate anymore, since the client construction happens in the `OkWrapper`.

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/CslbExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+import better.files.File
+import org.broadinstitute.monster.common.PipelineBuilderSpec
+
+class CslbExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
+  val outputDir = File.newTemporaryDirectory()
+  val cslbOutputDir = File(outputDir, CslbExtractionPipeline.subdir)
+  override def afterAll(): Unit = outputDir.delete()
+
+  // We know that some records were updated in this time range, so it should
+  // be fine to pull consistently without worrying about changes in data size.
```
ThePletch, line 14: this comment should be nixed, since we're not longer referencing live data

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -0,0 +1,76 @@
...
+import better.files.File
+import org.broadinstitute.monster.common.PipelineBuilderSpec
+
+class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
+  val outputDir = File.newTemporaryDirectory()
+  val hlesOutputDir = File(outputDir, HLESurveyExtractionPipeline.subdir)
+  override def afterAll(): Unit = outputDir.delete()
+
+  // We know that some records were updated in this time range, so it should
+  // be fine to pull consistently without worrying about changes in data size.
```
ThePletch, line 14: 🪓

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/CslbExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -0,0 +1,75 @@
...
+  val start =
+    OffsetDateTime.of(LocalDate.of(2020, 11, 15), LocalTime.MIDNIGHT, ZoneOffset.UTC)
+
+  val end =
+    OffsetDateTime.of(LocalDate.of(2020, 11, 16), LocalTime.MIDNIGHT, ZoneOffset.UTC)
```
ThePletch, line 19: also finding myself worrying that these timestamps are more specific than they need to be, since their value isn't relevant.

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/MockOkWrapper.scala
```diff
@@ -0,0 +1,54 @@
...
+
+  def makeRequest(request: Request): Future[Msg] = {
+    val p = Promise[Msg]()
+
+    val mapping = responseMap.map {
+      case (requestParams, mappedResponse) =>
+        RedCapClient.buildRequest(requestParams) -> mappedResponse
+    }
+
+    val matchingRequest = mapping.keys.find(requestKey => requestsMatch(requestKey, request))
```
ThePletch, line 43: this can probably be cleaned up to short circuit building every single request in the map, but since we're only usually doing two requests, i don't think it's a massive priority

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/RedcapMsgGenerator.scala
```diff
@@ -0,0 +1,20 @@
+package org.broadinstitute.monster.dap
+
+import upack._
+
+object RedcapMsgGenerator {
+
+  def redcapifyRecords(records: List[Map[String, String]]): Arr = {
```
ThePletch, line 7: not a huge fan of this function name, open to suggestions

ThePletch (commented):
self cr

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
```diff
@@ -81,25 +85,28 @@ class ExtractionPipelineBuilder(
...
           forms = formsForExtraction,
-          // List of fields to pull out of the data for us to filter on
           fields = fieldList
         )
       }
 
-    // Download the form data for each batch of records.
+    // Flatten the batches of results into a single, massive list once they've come back
     val extractedRecords = batchedIds.transform("Get records") {
       _.applyKvTransform(ParDo.of(lookupFn)).flatMap(kv => kv.getValue.fold(throw _, _.arr))
```
ThePletch, line 105: given how frequently this "grab batched results and flatten them to a list" line comes up, and how arcane it is, i think it'd benefit from being factored out into a common function, but i also don't think that's in scope for this PR.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/OkWrapper.scala
```diff
@@ -0,0 +1,42 @@
...
+          maybeResult match {
+            case Right(result) => p.success(result)
+            case Left(err)     => p.failure(err)
+          }
+        }
+      })
+    p.future
+  }
+
+  override def toString = "OkHTTP wrapper class"
```
aherbst-broad, line 41: ❓ Why override `toString` here? Clearer logging?

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/RedcapMsgGenerator.scala
aherbst-broad: `toRedcapFormat`?

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/RedcapMsgGenerator.scala
```diff
@@ -0,0 +1,24 @@
...
+      case (record, i) =>
+        record.map {
+          case (k, v) =>
+            Obj(
+              Str("record") -> Str(i.toString),
+              Str("field_name") -> Str(k),
+              Str("value") -> Str(v)
+            )
+        }
+    }: _*)
```
aherbst-broad, line 22: What's the `_*` operator doing in this context?

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/OkWrapper.scala
ThePletch: for some reason i was under the impression that it was a requirement to implement this for the `Serializable` trait, but presumably it can just use the default `Object.toString` implementation

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/RedcapMsgGenerator.scala
ThePletch: it's an expansion operator, since `Arr` seems to expect a series of elements as individual arguments rather than a single `Iterable`

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/CslbExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -0,0 +1,73 @@
+package org.broadinstitute.monster.dap
```
aherbst-broad, line 1: I (somewhat selfishly?) like that these are in just `test` now, not `it`, it should play well with Intellij's tooling better.

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/CslbExtractionPipelineBuilderIntegrationSpec.scala
aherbst-broad: Property based testing might be a thing we want to investigate, but I think these dates are fine for now.

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -0,0 +1,74 @@
...
+    )
+  )
+
+  override val builder =
+    HLESurveyExtractionPipeline.buildPipelineWithWrapper(mockWrapper)
+
+  behavior of "HLESurveyExtractionPipelineBuilder"
+
+  it should "successfully download records from RedCap" in {
+    readMsgs(hlesOutputDir, "records/*.json") shouldNot be(empty)
```
aherbst-broad, line 72: 🤔 Might be out of scope but: now that we control both sides of the interface with the mock wrapper, would it make sense to make assertions about the records themselves? i.e., more than just testing for not empty?

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/CslbExtractionPipelineBuilderIntegrationSpec.scala
ThePletch: the folder seems to affect `sbt` generally, which is why i made the move

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
ThePletch: yeah, we definitely need more robust testing than this. i'll make a ticket for increasing coverage before merging.

the main question i'd have is "what can we reasonably test about the records that isn't implicitly a test of redcap's API?"

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/MockOkWrapper.scala
```diff
@@ -0,0 +1,54 @@
+package org.broadinstitute.monster.dap
+
+import okhttp3._
+import upack.Msg
+
+import scala.concurrent.{Future, Promise}
+
+class MockOkWrapper(
+  responseMap: Map[RedcapRequestGeneratorParams, Msg] = Map.empty
```
aherbst-broad, line 9: 🤔 Would it be simpler to just take a list of responses and spit them out in order they were provided, rather than having to finding them in the mapping and having to do the object equality checks?

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilder.scala
```diff
@@ -81,25 +85,28 @@ class ExtractionPipelineBuilder(
     // NOTE: This logic is replicated from the encode-ingest pipeline,
     // we should consider moving it to scio-utils.
     val batchedIds = idsToExtract
+    // Pack up the IDs into dummy objects so beam knows how to handle them,
```
aherbst-broad, line 88: 👍 nice comment

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/MockOkWrapper.scala
ThePletch: i'm hesitant to regurgitate responses without verifying that we're responding to the expected request, since it could lead to hard to track down bugs or, worse, tests that pass despite the introduction of a request that will generate different data from the live API.

aherbst-broad (approved):
Had some comments, but lgtm overall.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/OkWrapper.scala
aherbst-broad: If this is Java's `Serializable` interface, it's only a marker interface/trait. I think the default `toString` should be ok.

raaidbroad (approved):

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/OkWrapper.scala
```diff
@@ -0,0 +1,40 @@
...
+import upack.Msg
+
+import scala.concurrent.{Future, Promise}
+
+class OkWrapper extends HttpWrapper {
+  /** Timeout to use for all requests to production RedCap. */
+  private val timeout = Duration.ofSeconds(60)
+
+  /** Construct a client instance backed by the production RedCap instance. */
+  val client = new OkHttpClient.Builder()
```
aherbst-broad, line 17: This causes a serialization error when running the pipeline via dataflow. If you make wrap it in a  `def` or `lazy val` that'll defer instantiation until first access in `makeRequest`

aherbst-broad (changes requested):
I'm bumping into a serialization error when testing this locally against @quazi-broad 's env branch, added a comment and potential fix.

aherbst-broad (approved):
LGTM, 🚢

PR #87, opened 2020-12-17
Configure fracture condition to allow description via spec field

ThePletch:
## Why

Fracture field is a checkbox, and has an optional description field to describe the nature of the fracture, which we had not been pulling in. This PR properly configures the field to pull in the spec field.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1330)

## This PR

### DIFF
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
@@ -22,6 +22,7 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
   )
 
   val subdir = "environment"
+
   //todo: need to query for all arms and work through arms serially
   val arm =
     List("baseline_arm_1", "dec2019_arm_1", "jan2020_arm_1")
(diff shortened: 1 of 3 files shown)


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -22,6 +22,7 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
   )
 
   val subdir = "environment"
+
```
quazi-broad, line 25: My IntelliJ keeps making this change for me and it has been so annoying.

quazi-broad (approved):
Change looks good to me, and unit test changes look good too!

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
ThePletch: i don't know that this is intelliJ, since i don't use it - i think this is an automated change by `scalafmt`

aherbst-broad (approved):
lgtm

raaidbroad (approved):

PR #88, opened 2020-12-18
Update RC variable being queried for stuffed toys

quazi-h:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1394)
DAP team noticed this field didn't have the correct value labels.
When I went to double check the transformation code I noticed we weren't even pulling the field out properly.

## This PR
-Updated the variable name being queried from a RawRecord to something that actually exists in RC.

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/RoutineEnvironmentTransformations.scala
@@ -217,7 +217,7 @@ object RoutineEnvironmentTransformations {
       dog.copy(
         deRoutineToys = toys,
         deRoutineToysIncludePlastic = rawRecord.getOptionalNumber("de_toy_plastic"),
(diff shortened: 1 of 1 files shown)


aherbst-broad (approved):

ThePletch (approved):
question to consider: should we change our variable names to match the format of theirs, i.e. change `deRoutineToysIncludeStuffedFabric` to `deRoutineToysIncludeFabricStuffed`? which way would be clearer?

quazi-broad:
> question to consider: should we change our variable names to match the format of theirs, i.e. change `deRoutineToysIncludeStuffedFabric` to `deRoutineToysIncludeFabricStuffed`? which way would be clearer?
Thanks Steve for that suggestion - I've created a catchall ticket so we don't forget about this inconsistency in the naming. I think that is definitely something we should do to minimize any confusion but also let's have this be low priority so we can catch any more as we get through the first data release - I've created a ticket here: https://broadinstitute.atlassian.net/browse/DSPDC-1420.

PR #90, opened 2020-12-18
Add okhttp logging interceptor

aherbst-broad:
## Why

We encounter errors with the DAP RedCAP API occasionally; it would be helpful to have the underlying http request + response when debugging in those circumstances.

## This PR
* OkHttp provides a [logging interceptor](https://github.com/square/okhttp/tree/master/okhttp-logging-interceptor). This PR adds a instance of it to the OkHttp client we're instantiating.
* Outputs the response body string when we cannot parse the response to JSON


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
```diff
@@ -46,6 +47,11 @@ object RedCapClient {
     val client = new OkHttpClient.Builder()
       .connectTimeout(timeout)
       .readTimeout(timeout)
+      .addInterceptor(new HttpLoggingInterceptor(new Logger() {
+        override def log(s: String): Unit = {
+          logger.info(s)
+        }
+      }).setLevel(Level.BASIC))
```
aherbst-broad, line 54: We can adjust this to also dump the headers + body when debugging; we're not doing that by default here to avoid accidentally outputting sensitive data to our prod logs (api tokens, etc.)

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -22,6 +22,7 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
   )
 
   val subdir = "environment"
+
```
ThePletch, line 25: getting a bit worried about the churn on this line - is this just because we haven't merged any of the multiple PRs that `scalafmt` made this change in, or will we continue to see diffs on this line in future PRs?

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
```diff
@@ -46,6 +47,11 @@ object RedCapClient {
     val client = new OkHttpClient.Builder()
       .connectTimeout(timeout)
       .readTimeout(timeout)
+      .addInterceptor(new HttpLoggingInterceptor(new Logger() {
+        override def log(s: String): Unit = {
```
ThePletch, line 51: scala syntactic sugar question - would it be possible to just directly overwrite this function, e.g. something like `override def log = logger.info`, since they have the same signature and (presumably) return type?

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
ThePletch: i'm surprised their levels have custom names like `BASIC` - how does this correspond to the standard log levels, e.g. `DEBUG`/`INFO`/`WARN`/`ERROR`?

ThePletch (approved):
changes overall look very straightforward - couple of concerns, but none that relate directly to the PR's fitness to merge

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
aherbst-broad: Yeah, I found this a bit surprising as well, the log levels are defined in [this enum](https://github.com/square/okhttp/blob/master/okhttp-logging-interceptor/src/main/kotlin/okhttp3/logging/HttpLoggingInterceptor.kt#L49). Short answer is: there isn't a 1-1 correspondence, you're more turning on parts of the HTTP request + response to capture.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/RedCapClient.scala
aherbst-broad: 🤔 That _seems_ like something scala might support, but I'm unaware of any syntactic sugar to that effect.

hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
aherbst-broad: Good eye, I think it's because we have yet to merge any of the multiple PRs.

quazi-h (approved):
LGTM. Thanks for catching that variable name update on the unit test!

PR #92, opened 2021-01-04
Add flag to parse and truncate floating point decimals in numeric field values

ThePletch:
## Why

We were getting values in `de_home_area` that included decimal points, which breaks Scala's integer parsing. We confirmed with the DAP team that it's safe to ignore decimal point values for integer fields.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1371)

## This PR

Adds a handler to truncate decimal values to `getOptionalNumber` in the raw record class. This handler can be manually disabled, but defaults to being on for all fields. The handler logs all decimals it finds as it truncates them. It does not swallow any errors other than a floating point value being passed to an integer field.

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
@@ -4,6 +4,8 @@ import java.lang.NumberFormatException
 import java.time.LocalDate
 import java.time.format.DateTimeFormatter
 
+import org.broadinstitute.monster.dap.HLESurveyTransformationPipelineBuilder.logger
+
 /**
   * Container for the raw properties pulled from RedCap for a single dog.
   *
@@ -51,7 +53,7 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
   def getOptionalBoolean(field: String): Option[Boolean] = getOptional(field).map(_ == "1")
 
   /** Get the singleton value for an attribute in this record, parsed as a long. */
-  def getOptionalNumber(field: String, truncateDecimals: Boolean = false): Option[Long] =
+  def getOptionalNumber(field: String, truncateDecimals: Boolean = true): Option[Long] =
     getOptional(field).map(value => {
       try {
         value.toLong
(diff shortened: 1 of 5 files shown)


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -50,7 +53,30 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
   def getOptionalBoolean(field: String): Option[Boolean] = getOptional(field).map(_ == "1")
 
   /** Get the singleton value for an attribute in this record, parsed as a long. */
-  def getOptionalNumber(field: String): Option[Long] = getOptional(field).map(_.toLong)
+  def getOptionalNumber(field: String, truncateDecimals: Boolean = true): Option[Long] =
+    getOptional(field).map(value => {
+      try {
+        value.toLong
```
aherbst-broad, line 59: 🔧 I think the python-ish EAFP pattern should be avoided in Java. Exceptions are expensive operations and should be avoided for control flow purposes. What if we check `truncateDecimals` first and if not set only then do the `toLong`? That way we can avoid the `try/catch` altogether and let the exception bubble up to the caller unchanged.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/RawRecord.scala
```diff
@@ -50,7 +52,24 @@ case class RawRecord(id: Long, fields: Map[String, Array[String]]) {
...
+  def getOptionalNumber(field: String, truncateDecimals: Boolean = true): Option[Long] =
+    getOptional(field).map(value => {
+      // we can presume the string represents a positive integer if all of its characters are decimal digits
+      if (truncateDecimals && !value.forall(Character.isDigit)) {
+        val truncatedValue: Long = value.toFloat.toLong
+
+        // don't log this error message until after we've successfully converted the string to a long.
+        // this avoids us logging this message erroneously if we were unable to parse the value,
+        // e.g. because it was garbled nonsense
+        TruncatedDecimalError(
```
aherbst-broad, line 64: OK, I'm a dope and didn't grok the full intent of this codepath (try _and_ log in the event of a bogus decimal value). In that case, I think the EAFP approach is the best way to go rather than the string parsing here, sorry about the confusion.

aherbst-broad (approved):
lgtm

quazi-h (approved):
The new TruncatedDecimalError looks good to me. I like that you are currently passing an argument to indicate when you want to use that logic. I think that by doing this way, each time we decide to make this conversion it will be very deliberate and we can also easily find where all we are practicing this if we need to look at it again later.

ThePletch:
@quazi-broad note that it currently defaults to true, since the ticket mentioned we were seeing multiple fields exhibit this behavior (which makes sense, since it sounds like this is caused by bad user input), so this behavior will apply to all numeric fields.

we can switch it to default to false in the future, if necessary, but i think it makes sense to explicitly enable the flag for any fields where we see this happen.

PR #93, opened 2021-01-04
Remove codecov config

aherbst-broad:
## Why

The code coverage tooling in this project are good for guidance but should not result in a failed check. 

## This PR
* Removes the codecov.yaml file which will revert us to the default codecov behavior, which is to not fail checks.

### DIFF
--- a/.github/codecov.yaml
+++ b/.github/codecov.yaml
@@ -1,7 +1,0 @@
-coverage:
-  status:
-    project:
-      default:
-        target: 75%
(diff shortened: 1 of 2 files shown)


ThePletch (approved):
:100: 🟢

quazi-h (approved):
LGTM, I think Steve's [PR](https://github.com/DataBiosphere/dog-aging-ingest/pull/90/files) includes the unit test variable name fix, too.

PR #96, opened 2021-01-07
Remove all zip code fields from transformation code and schema

ThePletch:
There should be no remaining references to ZIP codes in our code. A full refresh will be necessary to remove existing data from our databases.

## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1398)

The Redcap team has some privacy concerns about controlling access to ZIP code information, and since the level of granular control they need isn't currently available, they've asked us to simply remove all such fields from our tables and stop processing them in our transformations.

## This PR


aherbst-broad (approved):
Looks good.

quazi-h (approved):
Looks good to me! We can pair on you running a manual refresh once this is merged.

PR #97, opened 2021-01-08
Add support for reading directly from GCS to convert-output-to-tsv script

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1433)

#### PYTHON ENVIRONMENTS
* Adds a helper script (hack/run_in_virtualenv.sh) ported from ops' [Spitfire repo](https://github.com/broadinstitute/spitfire/) to let us run our Python scripts in environments with dependencies, Python versions, etc set up declaratively.

#### TSV CONVERSION SCRIPT
* Alters the convert-to-TSV script to read transformation outputs directly from GCS rather than needing to download them to the local filesystem first. Thanks to the `gcsfs` library, it'll still stream the file contents as it processes them, so this won't require either large downloads or high memory usage for large inputs.
* Updates script to introduce a few Python best practices (e.g. logging and argument parsing using core libraries) and simplify a few duplicated or otherwise unneeded branches of logic. Invocation syntax for marking the script as 'debug mode' will slightly differ, but otherwise this introduced no change in functionality.
* Adds support for converting only specified tables to TSVs rather than all tables by listing the desired tables in the invocation

#### MANUAL REFRESHES
* Operationalizes and checks in the ad-hoc manual refresh script to run with no user input from start to finish


hack/convert-output-to-tsv.py
```diff
@@ -71,94 +70,82 @@ def printd(x):
...
                     # store data
                     column_set.update(row.keys())
                     row_list.append(row)
                 else:
-                    print(f"Unrecognized table: {table_name}")
+                    log.info(f"Unrecognized table: {table_name}")
 
     # make sure pk is the first column
     # pop out the PK, will be splitting this set out later
     column_set.remove(entity_name)
```
ThePletch, line 89: i'm worried about this line because it's referring to a variable defined in a lower scope than this, so it's difficult to reason about what value it would have.

hack/convert-local-dir-to-tsv.py
```diff
@@ -0,0 +1,158 @@
+#!/usr/bin/env python
+
+import argparse
+import csv
+import json
+import logging
+from math import ceil
+import os
+
+from gcsfs.core import GCSFileSystem
```
ThePletch, line 10: dangit, this file should be the old version.

hack/convert-output-to-tsv.py
```diff
@@ -1,175 +1,157 @@
...
-                    else:
-                        print (f"Error, 'hs_condition_is_congenital' is not populated in {table_name}")
-                    row[entity_name] = ('%s-%s-%s' % 
-                        (row.get('dog_id'), row.get('hs_condition'), congenital_flag))
-                    # store data
-                    column_set.update(row.keys())
-                    row_list.append(row)
-                # environment:
+                    try:
+                        congenital_flag = int(congenital_flag)
```
aherbst-broad, line 79: 👍 much cleaner.

hack/run_in_virtualenv.sh
```diff
@@ -0,0 +1,125 @@
+#!/bin/bash
```
aherbst-broad, line 1: This may need a `chmod +x`

hack/run_in_virtualenv.sh
```diff
@@ -0,0 +1,125 @@
...
+            ;;
+        \? )
+            echo "Invalid option: -$OPTARG" 1>&2
+            print_help
+            exit 0
+            ;;
+    esac
+done
+
+ROOT_DIRECTORY=$(readlink -e $(dirname "$0"))
```
aherbst-broad, line 38: This bombs out for me when testing locally, with:
```
~/src/dog-aging-ingest/hack(sjp-dspdc-1433-improve-tsv-conversion-script*) » bash -x ./run_in_virtualenv.sh dap_venv                                                                                 (docker-desktop/default)
+ set -e
+ VERBOSE_MODE=0
+ getopts :vh OPTION
+++ dirname ./run_in_virtualenv.sh
++ readlink -e .
readlink: illegal option -- e
usage: readlink [-n] [file ...]
+ ROOT_DIRECTORY=
```
maybe an macos thing?

ThePletch: Ah, yeah, there's a helper I have to use the OS-specific `readlink` implementation depending on the OS running the command. Let me pull that in.

ThePletch: (`readlink` does something unrelated in OSX so you need to call `greadlink` to get the GNU functionality)

hack/manual_refresh.sh
```diff
@@ -0,0 +1,51 @@
...
+tmp_terra_tools=$(mktemp -d -t terratools-XXXXXXXX)
+git clone -q git@github.com:broadinstitute/terra-tools.git "$tmp_terra_tools"
+
+# Grab the Redcap tokens from Vault
+automation=$(vault read -field=token secret/dsde/monster/${ENV}/dog-aging/redcap-tokens/automation)
+env_automation=$(vault read -field=token secret/dsde/monster/${ENV}/dog-aging/redcap-tokens/env_automation)
+
+# EXTRACTION
+#   HLES
+# sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.HLESurveyExtractionPipeline --apiToken=$automation --pullDataDictionaries=false --outputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/raw --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=4 --maxNumWorkers=8 --experiments=shuffle_mode=service"
```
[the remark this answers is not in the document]
ThePletch: because i forgot to uncomment them after testing the transformation code, thanks for catching that

hack/convert-output-to-tsv.py
```diff
@@ -1,175 +1,157 @@
...
+
+    return text
+
+
+parser = argparse.ArgumentParser(description="Convert the records in a specified bucket prefix to a Terra-compatible TSV")
+parser.add_argument('input_dir', metavar='I', help='The bucket prefix to read records from')
+parser.add_argument('output_dir', metavar='O', help='The local directory to write the resulting TSVs to')
+parser.add_argument('table', nargs='*', help="One or more tables to process into TSVs. Processes all tables if none specified.")
+parser.add_argument('--debug', action='store_true', help="Write additional logs for debugging")
+parser.parse_args()
```
aherbst-broad, line 25: 🔧 I think you need to assign the parsing result to a var here, i.e., `args = parser.parse_args()`, and then use that var when pulling out the args like `args.table`, `args.input_dir`, etc.

ThePletch: yup, you're right. not sure how i missed that.

hack/convert-output-to-tsv.py
```diff
@@ -1,175 +1,157 @@
...
+        return text[len(prefix):]
+
+    return text
+
+
+parser = argparse.ArgumentParser(description="Convert the records in a specified bucket prefix to a Terra-compatible TSV")
+parser.add_argument('input_dir', metavar='I', help='The bucket prefix to read records from')
+parser.add_argument('output_dir', metavar='O', help='The local directory to write the resulting TSVs to')
+parser.add_argument('table', nargs='*', help="One or more tables to process into TSVs. Processes all tables if none specified.")
+parser.add_argument('--debug', action='store_true', help="Write additional logs for debugging")
```
aherbst-broad, line 24: ❓ Are you doing anything with this arg?

ThePletch: i have code locally to set the log level based on this but i have not committed it yet, will include in CR commit

hack/convert-output-to-tsv.py
```diff
@@ -1,175 +1,157 @@
...
-                    entity_name = pk_prefix + table_name + '_id'
-                    # env records are unique on dog_id, redcap_event
-                    # setting the primary key
                     dogId = row.get('dog_id')
                     eventName = row.get('address_month_year')
-                    row[entity_name] = ('%s-%s' %
-                        (dogId, eventName))
-                    column_set.update(row.keys())
-                    row_list.append(row)
+                    row[entity_name] = '-'.join([dogId, eventName])
```
aherbst-broad, line 86: I get a the following TypeError when testing this out:
`*** TypeError: sequence item 0: expected str instance, int found`

I believe dogId is an int being parsed as an int and will need to be cast to string.

aherbst-broad (changes requested):
Very very nice, thanks for cleaning this up. I ran into a few bumps when testing out locally, I think once those are ironed out this should be gtg.

aherbst-broad (approved):

quazi-h (approved):
Looks good, thanks for going over these changes with me during our pairing session!

PR #98, opened 2021-01-08
Strip newlines from cslb_other_changes

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1427)
Newlines in the `cslb_other_changes` field is causing issues with parsing by downstream systems. After an offline discussion, it's been determined that stripping the newlines will work for this field.

## This PR
* Adds a method to rawrecord that will return an `Option[String]` with newlines removed


raaidbroad (approved):

quazi-h (approved):
Looks pretty straightforward to me. I also realized we may be cleaning up any instances of this issue within the hles_dog data because of this line in the tsv script:
hack/convert-output-to-tsv.py

I wonder if we should remove that there after the initial data release to see if CRLF is getting into any of the hles_dog fields?

aherbst-broad:
@quazi-broad yeah, we should definitely remove it. I feel like this is 100% a transformation thing that we can do in Scala, it's strange to do it in Python (which we are not doing 100% of the time anyway, given this bug).

PR #99, opened 2021-01-12
Modified HealthTransformations to be able to process multiple health conditions per dog (bone fractures and rat poison).

quazi-broad:
…
## Why
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1372)
An HLES participant can include multiple options for the healthConditionOtherDescription field for two specific conditions (RatPoison and Fracture). Our current implementation does not allow for this and multiple of the same condition causes our transformation to fail.

## This PR
Adding an additional parameter to HealthCondition to indicate whether or not there are subcategories for that particular condition. We extended the RatPoison conditions to include one unique condition (with a unique health_condition_id) per TYPE of RatPoison.


ThePletch (approved):
Looks good on my end. The one worry I have is that the table rows generated by this solution aren't differentiated by anything other than their numeric value - their prefix will be the same for each one. Since our numbering scheme appears to be internal to our transformation code, I'd imagine that the DAP team will expect something more to identify which ID represents which subtype of fracture/rat poisoning.

aherbst-broad (approved):

PR #100, opened 2021-01-12
QH-DSPDC-1424-split-redcap-event

quazi-broad:
## Why
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1424)

## This PR
Added 3 columns to the Environment high level table (address month, year, and sequence).
Added logic to the EnvironmentTransformations code to parse out those three values from the RedcapEventName.

### DIFF
--- a/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
+++ b/hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
@@ -25,7 +25,6 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
 
   val arm =
     List(
-      "baseline_arm_1",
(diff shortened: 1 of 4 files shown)


hles/extraction/src/main/scala/org/broadinstitute/monster/dap/EnvironmentExtractionPipeline.scala
```diff
@@ -25,7 +25,6 @@ object EnvironmentExtractionPipeline extends ScioApp[Args] {
 
   val arm =
     List(
-      "baseline_arm_1",
```
aherbst-broad, line 28: Per a discussion with the DAP, we should not be pulling baseline_arm_1 data.

aherbst-broad (approved):
Caveat here is I worked a bit on this branch but: lgtm

ThePletch (approved):
approving this to get it into master

PR #102, opened 2021-01-12
Qh dspdc 1437 catch missing demographics

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1437)

There's a known bug with the data that the DAP team is sending us where it's missing calculated values for year and month for certain dogs. While we wait for them to fix this issue, we want to get out the records that aren't affected, so we need to change our code to skip over records that lack these required fields.

## This PR

* Makes our loggable exceptions extend `Throwable`
* Adds guards to calculated date-related fields in the dog demographics section that throws a relevant exception if they're missing. This is the only case where we want to actually throw an exception rather than silently log it, since record processing cannot proceed without these fields.
* Wraps dog transformation in a try-catch block to log and skip over records whose transformation is aborted.


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationLog.scala
```diff
@@ -6,7 +6,7 @@ import org.slf4j.Logger
 import ujson.Obj
 
 // a trait to capture the generic logging mechanism we'll want with case classes for the different logging levels
-abstract class HLESurveyTransformationLog {
+abstract class HLESurveyTransformationLog extends Throwable {
```
ThePletch, line 9: should we add a comment here saying "don't actually throw these exceptions unless the error requires the record to be skipped outright"?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -23,29 +23,29 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
     val rawRecords = readRecords(ctx, args)
     val rawEnvRecords = readEnvRecords(ctx, args)
 
-    val dogs = rawRecords.transform("Map Dogs")(_.map(DogTransformations.mapDog))
-    val owners = rawRecords.transform("Map Owners")(_.map(OwnerTransformations.mapOwner))
-    val cancer_conditions =
+    val dogs = rawRecords.transform("Map Dogs")(_.flatMap(DogTransformations.mapDog))
+    val owners = rawRecords.transform("Map Owners")(_.flatMap(OwnerTransformations.mapOwner))
+    val cancerConditions =
```
ThePletch, line 28: :camel:

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationPipelineBuilder.scala
```diff
@@ -23,29 +23,29 @@ object HLESurveyTransformationPipelineBuilder extends PipelineBuilder[Args] {
     val rawRecords = readRecords(ctx, args)
     val rawEnvRecords = readEnvRecords(ctx, args)
 
-    val dogs = rawRecords.transform("Map Dogs")(_.map(DogTransformations.mapDog))
-    val owners = rawRecords.transform("Map Owners")(_.map(OwnerTransformations.mapOwner))
-    val cancer_conditions =
+    val dogs = rawRecords.transform("Map Dogs")(_.flatMap(DogTransformations.mapDog))
+    val owners = rawRecords.transform("Map Owners")(_.flatMap(OwnerTransformations.mapOwner))
```
ThePletch, line 27: this change will mean that skipped records simply don't show up, rather than showing up as `null` like they did previously.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
```diff
@@ -56,10 +56,28 @@ object DemographicsTransformations {
   def mapAge(rawRecord: RawRecord, dog: HlesDogDemographics): HlesDogDemographics =
     rawRecord.getOptionalBoolean("dd_dog_birth_year_certain").fold(dog) { ageCertain =>
       if (ageCertain) {
-        val formYear = rawRecord.getRequired("dd_dog_current_year_calc").toInt
+        val dogId = rawRecord.id
+        val formYear = rawRecord.getOptional("dd_dog_current_year_calc") match {
+          case Some(year) => year.toInt
+          case None       =>
+            // we're throwing these errors instead of logging them because we want to completely skip records missing this value. they get logged
+            // by the logic up the chain that catches the error.
```
ThePletch, line 64: i don't like this comment duplication but i feel it's important to note right next to both places where it's happening

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
```diff
@@ -56,10 +56,28 @@ object DemographicsTransformations {
...
+              s"Record $dogId has less than 1 value for field dd_dog_current_year_calc"
+            )
+        }
+
         val birthYear = rawRecord.getRequired("dd_dog_birth_year").toInt
 
-        val formMonth = rawRecord.getRequired("dd_dog_current_month_calc").toInt
+        val formMonth = rawRecord.getOptional("dd_dog_current_month_calc") match {
+          case Some(month) => month.toInt
+          case None        =>
```
ThePletch, line 74: we tried to do this with a `map` and an `isEmpty` check, but this way ended up being much cleaner.

quazi-broad (approved):
Looks good to me - and we know it does work.
Thanks for getting this shipped!

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/HLESurveyTransformationLog.scala
aherbst-broad: +1 to a comment

aherbst-broad: Also, I think this should extend Exception, not Throwable

aherbst-broad (commented):
Looking good, can we get a test around this in DemographicsTransformationSpec?

aherbst-broad (approved):
lgtm

PR #103, opened 2021-01-12
Strip DOS and UNIX carriage returns

aherbst-broad:
## Why

In running through the cslb transform, I noticed we're still not removing offending newlines correctly. We need to strip DOS new lines (\r\n), not UNIX (\n). 

## This PR
* Changes the newline we replace to the DOS version (\r\n)


ThePletch (approved):
would it be overkill to use a regex or two replace statements to handle both flavors of line ending?

raaidbroad (approved):
Any chance we need to strip both `\n` and `\r\n`? If not, looks good.

aherbst-broad:
@ThePletch yeah good idea, updated.

ThePletch (approved):

PR #104, opened 2021-01-13
Limit HLES results to those from before 2021

ThePletch:
## Why

DAP has asked that the first release be all data from 2020.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1440)

## This PR

* Limit the pack date for records to before 2021


aherbst-broad (approved):
LGTM, it'd be good to verify locally against redcap prior to merging to make sure this does the desired filtering (i.e., we 100% do not see records from 2021)

quazi-broad (approved):

PR #105, opened 2021-01-13
Convert month to integer in environment transformations

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1442)

DAP team requested that month be represented as an integer rather than their bizarre inconsistent short-form month strings.

(In collaboration with Eric)

## This PR

* Convert month to integer in environment transformations


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
```diff
@@ -20,26 +21,48 @@ object EnvironmentTransformations {
...
+      case "mar"  => Some("3")
+      case "apr"  => Some("4")
+      case "may"  => Some("5")
+      case "jun"  => Some("6")
+      case "july" => Some("7")
+      case "aug"  => Some("8")
+      case "sept" => Some("9")
+      case "oct"  => Some("10")
+      case "nov"  => Some("11")
+      case "dec"  => Some("12")
```
ThePletch, line 36: i'm very worried by having discovered exactly two of these that are inexplicably four characters long instead of three. is there a place where i can see a list of all arms, for the sake of thoroughness?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
```diff
@@ -10,18 +11,58 @@ object EnvironmentTransformations {
...
+    }
+
+    // set address year
+    val addYear = redcapEventName(0).filter(_.isDigit)
+
+    addMonth match {
+      case None =>
+        val logger = LoggerFactory.getLogger(getClass)
+        val rawMonth = redcapEventName(0).filterNot(_.isDigit)
+        logger.error(s"Record ID ${dogId} has invalid raw month '${rawMonth}'")
```
aherbst-broad, line 47: You should increment [this counter](https://github.com/DataBiosphere/dog-aging-ingest/blob/master/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/PostProcess.scala#L7) so we get yelled at when the transformation finishes, otherwise this logging has a good chance of getting lost in the noise.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
quazi-broad: "baseline_arm_1", "dec2019_arm_1", "dec2019_secondary_arm_1", "jan2020_arm_1", "jan2020_secondary_arm_1", "feb2020_arm_1", "feb2020_secondary_arm_1", "mar2020_arm_1", "mar2020_secondary_arm_1", "apr2020_arm_1", "apr2020_secondary_arm_1", "may2020_arm_1", "may2020_secondary_arm_1", "june2020_arm_1", "june2020_secondary_arm_1", "july2020_arm_1", "july2020_secondary_arm_1", "aug2020_arm_1", "aug2020_secondary_arm_1", "sept2020_arm_1", "sept2020_secondary_arm_1", "oct2020_arm_1", "oct2020_secondary_arm_1", "nov2020_arm_1", "nov2020_secondary_arm_1", "dec2020_arm_1", "dec2020_secondary_arm_1"

aherbst-broad: We may able to do this via the RedCap environmental project console, I believe @quazi-broad has access.

aherbst-broad: @ThePletch alternatively there is an exportArms API call we can curl to get this data, we should be able to use that.

ThePletch: > `june2020_arm_1`
🙃🙃🙃🙃🙃🙃🙃

quazi-broad: I know python has a library that you can feed in a string such as the monthYear above and it will return a datetime. I wonder if scala has something similar.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/EnvironmentTransformationsSpec.scala
```diff
@@ -13,13 +13,18 @@ class EnvironmentTransformationsSpec extends AnyFlatSpec with Matchers with Opti
     val mapped = EnvironmentTransformations.mapEnvironment(
       RawRecord(
         1,
-        Map("redcap_event_name" -> Array("baseline_arm_1"), "baseline_complete" -> Array("2"))
+        Map(
```
quazi-broad, line 16: Just flagging that you will probably see a merge conflict here with some changes I pushed recently (removing baseline_arm_1 logic from env transformations)

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/EnvironmentTransformations.scala
ThePletch: we looked into using a datetime parsing library, but it ended up being harder to parse than just using a mapping like this.

quazi-broad (approved):
changes lgtm

aherbst-broad (approved):
:shipit:

PR #106, opened 2021-01-14
Use all address fields for environment table tsv primary key

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1443)

The DAP team has informed us that our current configuration for the environment table's primary key is not guaranteed to be unique.

## This PR

Adds the primary/secondary designation to the environment primary key in the generated TSV.


quazi-h (approved):
that should make that key unique

eparkmonster (approved):

PR #109, opened 2021-01-14
Qh dspdc 1445 inject puppy weight

quazi-h:
## Why
DAP has requested that we inject a "3" for records with a value of "2" under df_weight_change rather than passing that value through.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1445)

## This PR

Added a new case for when a value of 2 is passed into the gated question, mapping it to a 3 for the follow up response.
Also caught a minor bugfix - environment table fields where redcap_event_name was parsed into address1Or2, addressMonth, and addressYear should be integers in the Terra table. (This won't matter until we are in production and not using the firecloud uploads)

### DIFF
--- a/hles/logs/errors.log
+++ b/hles/logs/errors.log
@@ -1,0 +1,9 @@
+{"errorType":"MissingOwnerIdError","message":"Record 1 has less than 1 value for field st_owner_id"}
+{"errorType":"TruncatedDecimalError","message":"Record 1 has an unexpected decimal value in field de_home_area, truncated to integer"}
+{"errorType":"MissingOwnerIdError","message":"Record 33333 has less than 1 value for field st_owner_id"}
+{"errorType":"MissingOwnerIdError","message":"Record 11111 has less than 1 value for field st_owner_id"}
+{"errorType":"MissingOwnerIdError","message":"Record 33333 has less than 1 value for field st_owner_id"}
+{"errorType":"MissingOwnerIdError","message":"Record 11111 has less than 1 value for field st_owner_id"}
+{"errorType":"MissingOwnerIdError","message":"Record 11111 has less than 1 value for field st_owner_id"}
+{"errorType":"MissingOwnerIdError","message":"Record 33333 has less than 1 value for field st_owner_id"}
+{"errorType":"TruncatedDecimalError","message":"Record 1 has an unexpected decimal value in field cv_medianincomee, truncated to integer"}
--- a/hles/logs/errors.log
+++ b/hles/logs/errors.log
@@ -1,9 +1,0 @@
-{"errorType":"MissingOwnerIdError","message":"Record 1 has less than 1 value for field st_owner_id"}
-{"errorType":"TruncatedDecimalError","message":"Record 1 has an unexpected decimal value in field de_home_area, truncated to integer"}
-{"errorType":"MissingOwnerIdError","message":"Record 33333 has less than 1 value for field st_owner_id"}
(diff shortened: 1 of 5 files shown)


ThePletch:
man i hope nobody ever stumbles on this PR's title out of context

hles/logs/errors.log
```diff
@@ -0,0 +1,9 @@
+{"errorType":"MissingOwnerIdError","message":"Record 1 has less than 1 value for field st_owner_id"}
```
ThePletch, line 1: ⛔ let's not check this file in

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DietTransformations.scala
```diff
@@ -43,9 +43,12 @@ object DietTransformations {
       dfEverUnderweight = rawRecord.getOptionalNumber("df_underweight"),
       // Note: "df_overrweight" raw field name is misspelled in RedCap
       dfEverOverweight = rawRecord.getOptionalNumber("df_overrweight"),
+      // "1" from gated question maps to "yes, weight stayed the same" so we inject 0
+      // "2" from gated question maps to "puppy weight gain" so we inject a 3 to keep the distinction
       dfWeightChangeLastYear = weightChange.flatMap {
         case 1L    => Some(0L)
         case 0L    => rawRecord.getOptionalNumber("df_weight_change_how")
```
ThePletch, line 50: wouldn't `df_weight_change_how` be required if they answered `0L` here?

ThePletch (changes requested):
one thing that jumped out at me and one file to remove from the repo and possibly add to gitignore

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DietTransformations.scala
quazi-broad: Yup, that's not changing.
The options for the gated question are:
If we get a 1 we inject 0 into the follow up question's raw value
1 Yes
0 No
2 My dog is a puppy who was growing to his/her full size during the past year
99 Don't know

So we are correctly pulling the follow up when we hit 0. It's just that before this change we were passing through 2 as 2 for some reason. My thinking is that someone decided "My dog is a puppy who was growing to his/her full size during the past year" == "Gained Weight" in the follow up question - however DAP wants to keep the distinction by injecting "3" into the follow up response when we see a "2" in the gate.

hles/logs/errors.log
quazi-broad: yeah whoops, it should have pushed it out

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DietTransformations.scala
ThePletch: Sure, but I'm more referring to the use of `getOptionalNumber` rather than `getRequired` or similar for `df_weight_change_how` when the field is required if `df_weight_change` is answered with 0.

I don't think it's likely enough to fuss over too much, just something that jumped out at me.

quazi-broad: oh i see - yeah that's a good point.
actually we probably should enforce that, especially for these "gated question - follow up" scenarios. let me switch that to a getRequired and make sure everything compiles correctly

ThePletch (approved):

aherbst-broad (approved):

PR #110, opened 2021-01-15
Split transform pipelines by category so they can take different directories as inputs

ThePletch:
## Why

Our current transformation pipeline code attempts to process the data in the input directory as if it were all three types of raw data at once, which is leading to syntax and formatting errors.

## This PR

* Splits the pipeline logic into three different pipelines/pipeline builders so each type of data can be processed by a separate pipeline.


quazi-broad (commented):
Changes here looks good - if these would result in the same behavior as our extraction pipelines, I'm thinking you might need to make changes to the [manual_refresh.sh script](https://github.com/DataBiosphere/dog-aging-ingest/blob/master/hack/manual_refresh.sh#L58) to call the specific pipelines and also reflect the new subdirectories/paths being handled properly now

eparkmonster (approved):

quazi-broad (approved):
looks good to me!

PR #116, opened 2021-01-21
Add new columns for supplement gating questions

quazi-h:
## Why
The gating question for whether daily supplements or infrequent supplements are not in the hles_dog table.
As it stands now, a data user would have to check 10 other variables just to determine if daily supplements or infrequent supplements were given.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1462)

## This PR
- Added two new columns to hles_dog_diet json schema: df_daily_supplements, df_infrequent_supplements
- Added gating logic to the mapDailySupplements and mapInfrequentSupplements functions within DietTransformations
- Updated unit tests


aherbst-broad (approved):

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DietTransformations.scala
```diff
@@ -156,77 +156,96 @@ object DietTransformations {
...
-      dfDailySupplementsAcidify = rawRecord.getOptionalNumber("df_s_acidify").filter(_ != 0L),
-      dfDailySupplementsTaurine = rawRecord.getOptionalNumber("df_s_taurine").filter(_ != 0L),
-      dfDailySupplementsAntiox = rawRecord.getOptionalNumber("df_s_antiox").filter(_ != 0L),
-      dfDailySupplementsCoenzymeQ10 = rawRecord.getOptionalNumber("df_s_q10").filter(_ != 0L),
-      dfDailySupplementsOther = otherSupps,
-      dfDailySupplementsOtherDescription = if (otherSupps.nonEmpty) {
-        rawRecord.getOptional("df_s_other_text")
+  /** Map fields about supplements taken daily by a dog when available. */
+  def mapDailySupplements(rawRecord: RawRecord, dog: HlesDogDiet): HlesDogDiet =
+    rawRecord.getOptionalBoolean("df_supplement_daily").fold(dog) { dailySupps =>
```
ThePletch, line 161: we can probably turn this into a proper `if/else` statement instead of wrapping it in the `fold` by using `getRequired` instead of `getOptionalBoolean` here, since this field is required.

ThePletch (approved):
couple of suggestions about optional vs required (since the two yes/no questions here are both marked as 'required' in redcap), but otherwise looks good

PR #117, opened 2021-01-21
Parallelize manual refresh script to run all extract/transform  pipelines at once

ThePletch:
## Why

There's no reason for us to wait for each extraction/transformation pipeline to succeed to start the next one, since they're all independent operations.

## This PR
* Daemonizes each pipeline execution so all pipelines of the same type (extract or transform) run at once, sending their output to a file instead of the console to avoid interweaving messages.
* Wraps actions in functions with some boolean logic to ensure our script always cleans up after itself even if one of the pipelines fails.
* Adds some additional echo statements to make up for most of the script's output being redirected to files.


ThePletch:
n.b. this may not be doable, since sbt doesn't seem to support running multiple sbt processes at once. might be worth investigating if we can do this by submitting multiple simultaneous `runMain` commands to a single sbt invocation, which it explicitly does support.

PR #119, opened 2021-01-21
Removing duplicated address fields between two tables

quazi-h:
## Why
At some point there was a Data Model/Schema design decision that was made that resulted in us duplicating a set of 10 address columns across two of our tables: hles_dog and hles_owner. The DAP folks have reached out and asked if we can only include these columns in hles_owner as it is causing issues for their workflows to have them in both.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1470)

## This PR
Removed the 10 owner address columns from the hles_dog_residences json fragment, transformation code, and associated unit tests.
Added the two missing percentage fields to the owner table json, transformation code, and unit tests

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/OwnerTransformations.scala
@@ -70,6 +70,9 @@ object OwnerTransformations {
               } else {
                 None
               },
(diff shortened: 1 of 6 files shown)


aherbst-broad (approved):
Looks like a straightforward change.

ThePletch (approved):
what's the migration process for this, again? just run a full refresh and the tables will get picked up?

quazi-broad:
Yup, a full refresh should create new tables with the correct columns

PR #120, opened 2021-01-22
Update deprecated address fields for past dog residences

quazi-h:
## Why

Early in the study, the DAP folks were letting users enter their past residence countries with a free text field.
After they learned their lesson they implemented a drop down pick list. It appears we were never pulling the new dropdown fields so here is the logic to check both.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1457)

## This PR

- Added a new parseCountry method to check two separate fields - the deprecated free text field and then the drop down option.
- Added unit test and integration test cases.


aherbst-broad (approved):
lgtm

raaidbroad (approved):

PR #121, opened 2021-01-22
Fixed inconsistencies in activities parser

eparkmonster:
## Why
People may not have selected dd_activities_2 but selected dd_activities_m as a secondary activity. We want to keep the data for activities_m.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1473)
(https://broadinstitute.atlassian.net/browse/DSPDC-1474)

## This PR
Keeps the activity_m value even if the activity value wasn't selected. Another test was added to test this case

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
@@ -60,7 +60,7 @@ object DemographicsTransformations {
         val dogId = rawRecord.id
         val formYear = rawRecord.getOptional("dd_dog_current_year_calc") match {
           case Some(year) => year.toInt
-          case None       =>
+          case None =>
             // we're throwing these errors instead of logging them because we want to completely skip records missing this value. they get logged
             // by the logic up the chain that catches the error.
             throw MissingCalcFieldError(
@@ -72,7 +72,7 @@ object DemographicsTransformations {
 
         val formMonth = rawRecord.getOptional("dd_dog_current_month_calc") match {
           case Some(month) => month.toInt
-          case None        =>
+          case None =>
             // we're throwing these errors instead of logging them because we want to completely skip records missing this value. they get logged
             // by the logic up the chain that catches the error.
             throw MissingCalcFieldError(
@@ -223,7 +223,7 @@ object DemographicsTransformations {
     val country = usBorn match {
       case Some("1") => Some("US")
       case Some("0") => rawRecord.getOptional("dd_us_born_no")
-      case other     => other
+      case other => other
     }
     val source = rawRecord.getOptionalNumber("dd_acquire_source")
     val locationKnown = usBorn.contains("1") && rawRecord.getBoolean("dd_acquired_location_yn")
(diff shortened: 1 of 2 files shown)


hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
```diff
@@ -270,7 +270,7 @@ object DemographicsTransformations {
       if (allActivities.contains[Long](ActivityValues(activity))) {
         rawRecord.getOptionalNumber(s"dd_${activity}_m").orElse(Some(3L))
       } else {
-        None
```
aherbst-broad, line 273: I think a brief comment explaining this would be good

aherbst-broad (approved):

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
ThePletch: seconding the above, but i also think we can simplify this, since the only difference between these two branches is the value in `orElse`.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
```diff
@@ -270,7 +270,7 @@ object DemographicsTransformations {
...
+        rawRecord.getOptionalNumber(s"dd_${activity}_m").orElse(None)
```
ThePletch, line 273: worth noting, when simplifying this, that calling `orElse(None)` on an `Option[X]` is redundant, since it'll already be `None` if the `orElse` gets used.

hles/transformation/src/test/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformationsSpec.scala
```diff
@@ -532,5 +548,7 @@ class DemographicsTransformationsSpec extends AnyFlatSpec with Matchers with Opt
     assistanceOut.ddActivitiesServiceOther.value shouldBe true
     assistanceOut.ddActivitiesServiceOtherMedicalDescription.value shouldBe "Medical!"
     assistanceOut.ddActivitiesServiceOtherDescription.value shouldBe "Activity!"
+
+    activitiesOut.ddActivitiesCompanionAnimal.value shouldBe 1L
```
ThePletch, line 552: i think this spec is too big. when you test too many things like this, it becomes hard to identify what part of the service isn't working properly when the spec fails. can we split out the fallback behavior added in this PR into its own test?

ThePletch (approved):
there's a few things i think we can clarify/simplify here, but none that i think need to block us.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala
eparkmonster: Ok, reduced redundancy in this commit https://github.com/DataBiosphere/dog-aging-ingest/pull/121/commits/790e8361699b7d4970dd45fd031e918de5346b31

PR #122, opened 2021-01-27
Fix date-based filtering in CSLB by moving date filter back one second

ThePletch:
## Why

* The DAP team reported that CSLB records from January 1, 2021 were being included in our results, despite filtering to ensure only records from before 2021 were included.
* The date comparison logic in Redcap is bugged when comparing dates to a timestamp, such that a date evaluates as "less than" any timestamp on that date, including exactly midnight. This meant that filtering to all dates less than 2021-01-01 included all records from 2021-01-01, regardless of the specific time they were submitted (since that isn't recorded in CSLB).

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1480)

## This PR

* Fixes the above issue by adjusting our maximum timestamp from 2021-01-01 00:00:00 to 2020-12-31 23:59:59.
* In conclusion, PHP must be destroyed.


aherbst-broad (approved):

raaidbroad (approved):
lol

PR #123, opened 2021-01-28
Allow specifying state when 'US-born' is set to 'unknown'

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1494)

We were throwing away the state where a dog was acquired if their US-born status was "unknown", but this info was still being provided in that case and shouldn't be discarded.

## This PR


quazi-h (approved):
Looks good!

aherbst-broad (approved):
lgtm

PR #125, opened 2021-01-28
Do not filter out 0s for these values

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1497)
We were incorrectly suppressing the value for supplement variables if they were 0; we need to pass the 0s through.

## This PR
* Removes unneeded filtering 0 for supplement variables


quazi-h (approved):
lgtm

eparkmonster (approved):

PR #126, opened 2021-01-28
Un-negate 'without owner' dog activity questions to be 'with owner' instead

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1495)
* The DAP team requested that we update the "plays with humans without owner present" questions in the dog behavior survey to more closely match the actual wording in Redcap, where they are phrased as "with owner present."

## This PR
* Renames the fields in question from 'without' to 'with'
* Removes the negation we were applying to the boolean in question, to ensure that the resulting data's meaning is unchanged.


aherbst-broad (approved):
lgtm

raaidbroad (approved):

PR #132, opened 2021-01-29
Add data validation test

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1503)
We should have a repeatable validation test suite we can run over our data artifacts prior to delivery.

## This PR
* Implements a dead simple pytest suite that makes assertions over our dataset for the tickets we said are fixed in this week's release

### DIFF
--- a/hack/manual_refresh.sh
+++ b/hack/manual_refresh.sh
@@ -51,7 +51,7 @@ sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.HLESurveyE
 #   CSLB
 sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.CslbExtractionPipeline --apiToken=$automation --pullDataDictionaries=false --outputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/$REFRESH_SUBDIRECTORY/raw --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=4 --maxNumWorkers=8 --experiments=shuffle_mode=service --endTime=2020-12-31T23:59:59-05:00"
 #   ENVIRONMENT
-sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.EnvironmentExtractionPipeline --apiToken=$env_automation --pullDataDictionaries=false --outputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/$REFRESH_SUBDIRECTORY/raw --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=4 --maxNumWorkers=8 --experiments=shuffle_mode=service --endTime=2020-12-31T23:59:59-05:00"
+#sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.EnvironmentExtractionPipeline --apiToken=$env_automation --pullDataDictionaries=false --outputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/$REFRESH_SUBDIRECTORY/raw --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=4 --maxNumWorkers=8 --experiments=shuffle_mode=service --endTime=2020-12-31T23:59:59-05:00"
 
 # TRANSFORMATION
 #   HLES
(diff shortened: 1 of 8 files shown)


hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+        reader = csv.DictReader(f, delimiter='\t')
+        for row in reader:
+            rows.append(row)
+
+    return rows
+
+
+def test_cslb_date(cslb_data):
+    for row in cslb_data:
+        assert '2021' not in row['cslb_date']
```
ThePletch, line 36: not sure how much effort we want to put into this validation, but there's two things we can do to make failures easier to debug:

1. use unittest as our testing library (we can still run via pytest) and wrap these tests in a class that extends `unittest.TestCase` so we can use unittest's assertion methods, e.g. `self.assertNotIn('2021', row['cslb_date'])`, which produce failure messages with more context, rather than `assert`, which just raises an `AssertionError` saying that the assertion was false.
2. add some info about the current row to the failure message so we get a more actionable error message than `'2021' found in '2021-02-02'`

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+        assert '2020' in row['cslb_date']
+
+
+def test_spay_or_neuter_age(hles_dog2_data):
+    for row in hles_dog2_data:
+        assert 'dd_spay_or_neuter_age' in row
+        assert row['dd_spay_or_neuter_age'] in {'', '5', '99', '2', '4', '7', '6', '8', '3', '1'}
```
ThePletch, line 43: can we sort these values? it's not particularly obvious to me at a glance that this list includes the numbers 1 through 8.

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_dd_acquired_st(hles_dog1_data):
+    found_affected_row = False
+    for row in hles_dog1_data:
+        if row['entity:dog_id_id'] in ('18069', '32705', '40519', '63438', '89055'):
+            found_affected_row = True
+            assert row['dd_acquired_state'] not in {'', 'NA'}
+
+    assert found_affected_row is True
```
ThePletch, line 97: extreme nitpick alert
```suggestion
    assert found_affected_row
```

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_dd_activities_obedience(hles_dog1_data):
+    found_affected_row = False
+    for row in hles_dog1_data:
+        if row['entity:dog_id_id'] in {'76099'}:
+            assert 'dd_activities_obedience' in row
+            assert row['dd_activities_obedience'] == ''
+            found_affected_row = True
+    assert found_affected_row is True
```
ThePletch, line 107:
```suggestion
    affected_obedience_row = next(row for row in hles_dog1_data if row['entity:dog_id_id'] == '76099')
    assert 'dd_activities_obedience' in row
    assert row['dd_activities_obedience'] == ''
```
this will raise a `StopIterationError` if no such row is found

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_dd_activities_agility(hles_dog1_data):
+    found_affected_row = False
+    for row in hles_dog1_data:
+        if row['entity:dog_id_id'] in {'34888'}:
+            assert 'dd_activities_agility' in row
+            assert row['dd_activities_agility'] == ''
+            found_affected_row = True
+    assert found_affected_row is True
```
ThePletch, line 117: see above re: looking up single row IDs

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+            assert 'dd_activities_working' in row
+            assert row['dd_activities_working'] == ''
+            found_affected_row = True
+    assert found_affected_row is True
+
+
+def test_st_portal_account_creation_date(hles_dog1_data):
+    for row in hles_dog1_data:
+        assert 'st_portal_account_creation_date' in row
+        assert len(row['st_portal_account_creation_date']) > 0
```
ThePletch, line 163:
```suggestion
        assert bool(row['st_portal_account_creation_date'])
```
up to you if this sacrifice in readability is acceptable

hack/validation/test_release.py
```diff
@@ -0,0 +1,181 @@
...
+
+
+def test_past_residence_country(hles_dog2_data, hles_dog1_data):
+    past_residence_country1_vals = set()
+    past_residence_country2_vals = set()
+
+    for row in hles_dog2_data:
+        assert 'de_past_residence_country1_text' in row
+        if row['de_past_residence_country1_text']:
+            past_residence_country1_vals.add(row['de_past_residence_country1_text'])
```
ThePletch, line 173:
```suggestion
    assert any(row['de_past_residence_country1_text'] for row in hles_dog2_data)
```

ThePletch (approved):
couple of suggestions for python core libraries to pull in and some syntactic sugar/readability suggestions

hack/manual_refresh.sh
```diff
@@ -51,28 +51,28 @@ sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.HLESurveyE
...
+#sbt "dog-aging-hles-transformation/runMain org.broadinstitute.monster.dap.EnvironmentTransformationPipeline --inputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/$REFRESH_SUBDIRECTORY/raw/environment --outputPrefix=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/$REFRESH_SUBDIRECTORY/transform --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=6 --maxNumWorkers=10 --experiments=shuffle_mode=service --dumpHeapOnOOM --saveHeapDumpsToGcsPath=gs://broad-dsp-monster-dap-dev-storage/weekly_refresh/transform/OOMlogs"
```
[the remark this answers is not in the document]
aherbst-broad: 🤦 yup, will fix.

raaidbroad (approved):
+1 to Steve's comments, otherwise looks good. unittest does provide better context but idk how much effort we want here

hack/validation/test_release.py
aherbst-broad: I'll shift to unittest and add the descriptive msg, good call.

PR #134, opened 2021-01-29
Add new columns for past country residence

quazi-h:
## Why
DAP has 10 fields to get the alternate past residence country for a participant.
They initially were allowing users to enter their own text strings and have since deprecated that field and are now providing a dropdown to choose a country form. In our old logic we were checking both. DAP has requested that we add 10 columns for the "text entry" past residences.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1484)

## This PR
- Added 10 new fields to the ResidentialEnvironmentTransformations
- Modified the parseCountry to beck for both the  "_01" vs "_01_only" RC variables
- Updated unit tests

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
@@ -28,12 +28,12 @@ object ResidentialEnvironmentTransformations {
     * de_country_[n] is the deprecated user entry free text field
     * de_country_[n]_dd is the updated drop down option list
     */
-  def parseCountry(
+  def parseOnlyCountry(
     rawRecord: RawRecord,
-    countryField: String,
-    countryFieldDropdown: String
+    multipleAdd: String,
+    only1Add: String
   ): Option[String] = {
-    rawRecord.getOptional(countryField).orElse(rawRecord.getOptional(countryFieldDropdown))
+    rawRecord.getOptional(multipleAdd).orElse(rawRecord.getOptional(only1Add))
   }
 
   /**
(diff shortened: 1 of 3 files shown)


quazi-broad:
Two things:

- Drew made some updates on his branch that need to be pushed.
- The logic in transformations could be refined - may do this Monday

Somewhat irrelevant to the PR but won't adding 10 new columns for data that only a handful of records have cost more than its worth? (not complaining - just thinking that is a little overkill)

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -28,12 +28,12 @@ object ResidentialEnvironmentTransformations {
...
-  def parseCountry(
+  def parseOnlyCountry(
     rawRecord: RawRecord,
-    countryField: String,
-    countryFieldDropdown: String
+    multipleAdd: String,
+    only1Add: String
   ): Option[String] = {
-    rawRecord.getOptional(countryField).orElse(rawRecord.getOptional(countryFieldDropdown))
+    rawRecord.getOptional(multipleAdd).orElse(rawRecord.getOptional(only1Add))
```
ThePletch, line 36: it's a little unclear to me how the function name and argument names relate. specifically, a function that parses both fields but is named `parseOnlyCountry` is confusing to me.

could we either change these names or add comments clarifying what each argument of the function is?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -58,38 +58,54 @@ object ResidentialEnvironmentTransformations {
...
-          dePastResidenceCountry2 =
-            if (pastCountryCount > 1) parseCountry(rawRecord, "de_country_02", "de_country_02_dd")
+          dePastResidenceCountry1 =
+            if (pastCountryCount > 1)
+              rawRecord
+                .getOptional("de_country_01_dd")
+                .orElse(rawRecord.getOptional("de_country_01_only_dd"))
             else None,
+          dePastResidenceCountry2 =
+            if (pastCountryCount > 2) rawRecord.getOptional("de_country_02_dd") else None,
```
ThePletch, line 68: rather than copy-paste this line eighteen times, let's make a function for it and use that in all eighteen places instead.

ThePletch (changes requested):
couple of code cleanup suggestions to improve clarity and reduce repetition.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
quazi-broad: Added comments to the function to explain the new args

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
quazi-broad: I renamed the function and moved the if (count check) and else None into the function.
I feel like this could be even further simplified if we just iterated over a range [1-10] to check for a certain prefix + suffix of RC variable for a given column in our table, something like:
`dePastResidenceCountry[N]        = getPastResidenceCountry(prefix="de_country_",suffix="_dd",range=2-10)
dePastResidenceCountry[N]Text = getPastResidenceCountry(prefix="de_country_",suffix=None,range=2-10)`
but I think what we have now may be better because its easier to read and understand the RC variable names we are checking. Let me know your thoughts

aherbst-broad (approved):

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
ThePletch: i think abstracting over the prefix/suffix would be overkill, especially since we'd need to manually assign each one to its specific attribute anyway.

ThePletch (approved):

PR #135, opened 2021-02-03
Move TSV generation into the transformation pipelines

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1444)

* Our existing Python script to transform JSON outputs into TSVs is difficult to maintain and test, and doesn't take advantage of any dataflow constructs.

## This PR

* Adds support for generating TSVs directly from our transform steps' JSON outputs, including dynamically generating headers from our Jade table case classes
* Adds tests for the utilities used to do so
* Standardizes our transform pipelines to reduce repetition when doing these same steps for each category of data

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
@@ -1,0 +1,41 @@
+package org.broadinstitute.monster.dap
+
+import scala.reflect.runtime.universe.{Symbol, Type, TypeTag, typeOf, termNames}
+
+object CaseClassInspector {
+
+  def snakeCaseHeaderRow[T: TypeTag](delimiter: String = "\t"): String =
+    snakeCaseHeaderList[T].mkString(delimiter)
+
+  def snakeCaseHeaderList[T: TypeTag]: Seq[String] = snakeCaseHeaderList(typeOf[T])
+
+  private def snakeCaseHeaderList(t: Type) =
+    flattenParamsList(t).map(_.name.toString).map(snakify(_))
+
+  private def flattenParamsList(t: Type): Seq[Symbol] = {
+    caseClassParamsList(t).flatMap(expandParamsList(_))
+  }
+
+  private def caseClassParamsList(t: Type): Seq[Symbol] =
+    t.decl(termNames.CONSTRUCTOR).asMethod.paramLists(0)
+
+  private def expandParamsList(field: Symbol): Seq[Symbol] = {
+    getNestedSymbols(field.info).getOrElse(Seq(field))
+  }
+
+  // courtesy of https://github.com/lift/framework/blob/master/core/util/src/main/scala/net/liftweb/util/StringHelpers.scala
+  private def snakify(name: String) =
+    name
+      .replaceAll("([A-Z]+)([A-Z][a-z])", "$1_$2")
+      .replaceAll("([a-z\\d])([A-Z])", "$1_$2")
+      .replaceAll("([a-z])(\\d)", "$1_$2")
+      .toLowerCase
+
+  private def getNestedSymbols(targetType: Type): Option[Seq[Symbol]] = {
+    targetType match {
+      case optional if optional <:< typeOf[Option[Any]] => getNestedSymbols(optional.typeArgs(0))
+      case product if product <:< typeOf[Product]       => Some(flattenParamsList(product))
+      case _                                            => None
+    }
+  }
+}
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/ScioUtils.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/ScioUtils.scala
@@ -1,0 +1,39 @@
+package org.broadinstitute.monster.dap
+
+import com.spotify.scio.values.SCollection
+
+import io.circe.syntax._
+import io.circe.{JsonObject, Encoder}
+
+import org.broadinstitute.monster.common.StorageIO
+
+object ScioUtils {
+
+  def jsonSerializeTransform[M: Encoder](messages: SCollection[M], description: String) = {
+    messages.transform(s"Convert '$description' messages to JSON")(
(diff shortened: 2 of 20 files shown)


ThePletch:
I'm not very happy with the test coverage for the changes to the pipelines themselves, but I feel like adding more in-depth integration test coverage is out of scope for this PR. I've confirmed that the JSON outputs are unchanged and the TSV outputs (for the sample records I pulled out) match up with our Python script's outputs other than the order of the columns.

project/plugins.sbt
```diff
@@ -10,5 +10,6 @@ resolvers += Resolver.url(
   "Broad Artifactory",
   new URL("https://broadinstitute.jfrog.io/broadinstitute/libs-release/")
 )(publishPatterns)
+resolvers += Resolver.sonatypeRepo("releases")
```
ThePletch, line 13: i think we can remove this since i didn't end up using the library that required this

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/TsvUtils.scala
```diff
@@ -0,0 +1,70 @@
...
+    val writer = new CSVWriter(rowBuilder, '\t', '"', '"', "\n")
+    // the second argument tells the writer how to handle quoting.
+    // if it's false, it will only wrap cells containing quote/newline/separator characters in quotes.
+    writer.writeNext(cells.toArray, false)
+
+    rowBuilder.toString
+  }
+
+  // presumes that the element is not itself a json object (i.e. json objects we map this function across must be flat)
+  def jsonElementToString(element: Json): String = {
```
aherbst-broad, line 44: ⛏️ Would it make sense to add a `case _` to catch and throw in circumstances where this function is passed an object?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
```diff
@@ -0,0 +1,41 @@
...
+  private def snakify(name: String) =
+    name
+      .replaceAll("([A-Z]+)([A-Z][a-z])", "$1_$2")
+      .replaceAll("([a-z\\d])([A-Z])", "$1_$2")
+      .replaceAll("([a-z])(\\d)", "$1_$2")
+      .toLowerCase
+
+  private def getNestedSymbols(targetType: Type): Option[Seq[Symbol]] = {
+    targetType match {
+      case optional if optional <:< typeOf[Option[Any]] => getNestedSymbols(optional.typeArgs(0))
```
aherbst-broad, line 36: ❓ What's the `<:<` operator doing here?

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/ScioUtils.scala
```diff
@@ -0,0 +1,39 @@
...
+
+object ScioUtils {
+
+  def jsonSerializeTransform[M: Encoder](messages: SCollection[M], description: String) = {
+    messages.transform(s"Convert '$description' messages to JSON")(
+      _.map((msg: M) => msg.asJson.printWith(StorageIO.circePrinter))
+    )
+  }
+
+  def toJsonAndTsv[M: Encoder](
```
aherbst-broad, line 18: 🤔 I'm wondering if this should consume a `numShards` argument that we can pass in to `saveAsTextFile` so we can end up with a single TSV rather than having to stitch together >=1 after the job.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
ThePletch: this is saying "if this class can be cast to `Option[Any]`", i.e. "is it an Option[X] type"

ThePletch: since you can't do pattern matching on types, to my knowledge

aherbst-broad:
I believe there are a handful of fields whose header is not strictly snake case, i.e., `de_past_residence_country2` in the current "hack" tsv output vs `de_past_residence_country_2` in the new output. I am unsure why that's the case (perhaps how python renders tsv headers?). The new headers should line up with the "hack" version if at all possible to avoid impacting downstream.

ThePletch:
frustratingly, we've got it both ways: sometimes we have `de_past_residence_country1` and sometimes we have `tp_pcpn_norm_01`. i'm digging into our headers to see what distinguishes these/why they render differently in the JSON outputs.

ThePletch:
looks like the cause of this distinction is that the names of the fields in the jade schema are inconsistently rendered (as we've observed, sometimes they separate by underscores, sometimes they don't). presumably our json serializer has some way of accessing the spelling used in the jade schema json. trying to find where/how it does that now.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
raaidbroad: You should be able to pattern match on a type: https://docs.scala-lang.org/tour/pattern-matching.html (sorry the page doesn't anchor link, if you search for "Matching on type only" it should jump to the right section)

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/ScioUtils.scala
raaidbroad: +1

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/CaseClassInspector.scala
ThePletch: @raaidbroad sadly not, that only lets you match if a value is an instance of the given type. these are actual type objects, so this is essentially matching on value.

ThePletch:
Closing following decision not to pursue solutions involving Shapeless. Looking like this isn't very tenable without significantly more work, which we're not willing to dedicate at this time.

PR #137, opened 2021-02-04
Expanded integration tests for DAP extraction pipeline

eparkmonster:
## Why
Our integration tests for DAP extraction pipeline did not extend beyond checking if the json file downloaded was not empty

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1419)

## This PR
Adds tests to make sure that each record has a record, field_name, and value key and a test to check expected fields with their expected values

### DIFF
--- a/hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
+++ b/hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
@@ -25,7 +25,8 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
       pullDataDictionaries = false
     )
 
-  val expectedStudyIds = List("id_one", "id_two")
+  val fakeIds = 0 to 1
+  val expectedStudyIds = fakeIds.map { id => s"id_${id}" }.toList
 
   val studyIdsRequest = RedcapRequestGeneratorParams(
     testArgs.apiToken,
@@ -69,8 +70,6 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
   it should "successfully download records from RedCap" in {
     readMsgs(hlesOutputDir, "records/*.json") shouldNot be(empty)
   }
-
-  val fakeIds = 0 to 1
 
   val expectedPackDateRecord: Seq[Obj] = fakeIds.map { i =>
     Obj(
--- a/hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
+++ b/hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
@@ -3,6 +3,7 @@ package org.broadinstitute.monster.dap
 import java.time.{LocalDate, LocalTime, OffsetDateTime, ZoneOffset}
 import better.files.File
 import org.broadinstitute.monster.common.PipelineBuilderSpec
+import upack._
 
 class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[Args] {
   val outputDir = File.newTemporaryDirectory()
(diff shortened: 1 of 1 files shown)


hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -69,4 +68,28 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
...
+      val record = msg.obj.toString()
+      record.contains("record") shouldBe true
+      record.contains("field_name") shouldBe true
+      record.contains("value") shouldBe true
+    })
+  }
+
+  it should "contain records with expected valid fields" in {
+    val allRecords = (readMsgs(hlesOutputDir, "records/*.json")).toString()
+    print(allRecords)
```
aherbst-broad, line 83: 🔧 Remove this print

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -69,4 +68,28 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
   it should "successfully download records from RedCap" in {
     readMsgs(hlesOutputDir, "records/*.json") shouldNot be(empty)
   }
+
+  it should "contain records that all have a record, field_name, and value key" in {
+    readMsgs(hlesOutputDir, "records/*.json").foreach(msg => {
+      val record = msg.obj.toString()
```
aherbst-broad, line 74: 🤔 I think we can make some stronger assertions here by not deserializing to a string but rather leaving the message as an object and comparing to an expected object, much like we do in the ExtractionPipelineBuilderSpec [here](https://github.com/DataBiosphere/dog-aging-ingest/blob/e3deb8a07bab3b9febc8a6f73f030856a434bd39/hles/extraction/src/test/scala/org/broadinstitute/monster/dap/ExtractionPipelineBuilderSpec.scala#L37-L41)

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -69,4 +68,28 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
...
+      record.contains("record") shouldBe true
+      record.contains("field_name") shouldBe true
+      record.contains("value") shouldBe true
+    })
+  }
+
+  it should "contain records with expected valid fields" in {
+    val allRecords = (readMsgs(hlesOutputDir, "records/*.json")).toString()
+    print(allRecords)
+    allRecords.contains("record") shouldBe true
```
aherbst-broad, line 84: Same comment about not deserializing to string and instead comparing to a comparision upack `Obj`

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
ThePletch: +1 to this, directly accessing keys and values avoids accidental false positives from the string we're looking for showing up in an unexpected part of the string, but not in the field we want to check

ThePletch (dismissed):
agreeing with drew's comments. let me know if you want to pair on digging directly into the messages here - i know upack's typing system is pretty opaque.

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
eparkmonster: Tests have been changed to compare to the expected object and to compare the fields and values that we are looking for

hles/extraction/src/test/scala/org/broadinstitute/monster/dap/HLESExtractionPipelineBuilderIntegrationSpec.scala
```diff
@@ -69,4 +69,37 @@ class HLESExtractionPipelineBuilderIntegrationSpec extends PipelineBuilderSpec[A
   it should "successfully download records from RedCap" in {
     readMsgs(hlesOutputDir, "records/*.json") shouldNot be(empty)
   }
+
+  val fakeIds = 0 to 1
```
aherbst-broad, line 73: It'd be good to tie this with the expectedStudyIds val we are declaring on line 28, so we can avoid future test breakage if we change the set of IDs. What if we generate the expectedStudyIds val from this list?

aherbst-broad (approved):

aherbst-broad (approved):
LGTM

quazi-h (approved):
This looks good to me - I see inputs for the 3 RC variables that should get checked.
Approving

PR #138, opened 2021-02-04
update logic and tests for single past country

raaidbroad:
## Why
Looks like there is a small bug in what country is pulled if there is only one other country of residence prior to the current residence (see relevant ticket).

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1521?atlOrigin=eyJpIjoiZDFlM2Y3ZDUwOWYxNDhmYmI0NGMyOWIxYTZmYTkyMzMiLCJwIjoiaiJ9)

## This PR
Attempts to fix said issue and update tests to enforce correct behavior.

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
@@ -64,11 +64,9 @@ object ResidentialEnvironmentTransformations {
           dePastResidenceCountryCount = Some(pastCountryCount),
           // there are two separate fields to check for "First country"
           dePastResidenceCountry1 =
-            if (pastCountryCount > 1)
-              rawRecord.getOptional("de_country_01_dd")
-            else if (pastCountryCount == 1)
+            if (pastCountryCount == 1)
               rawRecord.getOptional("de_country_01_only_dd")
-            else None,
+            else getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_01_dd"),
           dePastResidenceCountry2 =
             getPastResidenceCountry(rawRecord, pastCountryCount, 2L, "de_country_02_dd"),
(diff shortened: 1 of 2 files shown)


aherbst-broad (approved):
This looks ok, @quazi-broad should take a look to ensure the business logic is sound.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -65,9 +65,9 @@ object ResidentialEnvironmentTransformations {
           // there are two separate fields to check for "First country"
           dePastResidenceCountry1 =
             if (pastCountryCount > 1)
```
ThePletch, line 67: it won't let me make a suggestion that spans deleted lines, so don't click auto-accept, but this is my suggestion for simplifying this field's logic

```suggestion
          dePastResidenceCountry1 =
            if (pastCountryCount == 1) rawRecord.getOptional("de_country_01_only_dd") else getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_01_dd")
```

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -92,7 +92,8 @@ object ResidentialEnvironmentTransformations {
...
               rawRecord
                 .getOptional("de_country_01")
-                .orElse(rawRecord.getOptional("de_country_01_only"))
+            else if (pastCountryCount == 1)
+              rawRecord.getOptional("de_country_01_only")
```
ThePletch, line 96: see above - i think we can simplify this block in the same way

ThePletch (approved):
logic is correct, but one suggestion to improve readability/brevity

aherbst-broad (approved):

PR #141, opened 2021-02-05
Update country counter integers

quazi-h:
## Why
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1507)
With our most recent update to this block of code, the integer we are passing in to the _getPastResidenceCountry_ was off by 1.

## This PR
- Incrementing the pastCountryCount by 1 (bug).
- Updated unit tests to test this specific logic.

### DIFF
--- a/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
+++ b/hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
@@ -66,27 +66,27 @@ object ResidentialEnvironmentTransformations {
           dePastResidenceCountry1 =
             if (pastCountryCount > 1)
               rawRecord
-                .getOptional("de_country_01_dd")
+                .getOptional("de_country_deLifetimeResidenceCount01_dd")
                 .orElse(rawRecord.getOptional("de_country_01_only_dd"))
             else None,
           dePastResidenceCountry2 =
-            getPastResidenceCountry(rawRecord, pastCountryCount, 2L, "de_country_02_dd"),
+            getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_02_dd"),
           dePastResidenceCountry3 =
-            getPastResidenceCountry(rawRecord, pastCountryCount, 3L, "de_country_03_dd"),
+            getPastResidenceCountry(rawRecord, pastCountryCount, 2L, "de_country_03_dd"),
           dePastResidenceCountry4 =
-            getPastResidenceCountry(rawRecord, pastCountryCount, 4L, "de_country_04_dd"),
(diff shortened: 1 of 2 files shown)


aherbst-broad (approved):
lgtm ,we cowboyed this last week manually and it worked.

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -68,47 +68,46 @@ object ResidentialEnvironmentTransformations {
               rawRecord.getOptional("de_country_01_only_dd")
             else getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_01_dd"),
           dePastResidenceCountry2 =
-            getPastResidenceCountry(rawRecord, pastCountryCount, 2L, "de_country_02_dd"),
+            getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_02_dd"),
```
raaidbroad, line 71: Line 69 above here also uses `1L`; should that then be 0??

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
```diff
@@ -68,47 +68,46 @@ object ResidentialEnvironmentTransformations {
...
           // there are two separate fields to check for "First country"
           dePastResidenceCountry1Text =
             if (pastCountryCount == 1)
-              rawRecord
-                .getOptional("de_country_01_only")
+              rawRecord.getOptional("de_country_01_only")
             else getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_01"),
           dePastResidenceCountry2Text =
-            getPastResidenceCountry(rawRecord, pastCountryCount, 2L, "de_country_02"),
+            getPastResidenceCountry(rawRecord, pastCountryCount, 1L, "de_country_02"),
```
raaidbroad, line 94: Line 92 above also uses `1L`; should that be 0??

hles/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/ResidentialEnvironmentTransformations.scala
quazi-broad: No I think that's good.
Here's the original logic from before recent changes:
https://github.com/DataBiosphere/dog-aging-ingest/commit/384e8b0d89aad198a1312fe6ad7f30ffbe337d7b

Was checking for >1 before because the "only" fields indicate only one country, otherwise there should be more than 1. Good catch!

raaidbroad (approved):

PR #142, opened 2021-02-05
Added new TSV validations

quazi-h:
## Why
We got an updated list of issues with the HLES dataset which we were able to address with code changes.
After creating the TSV files, we run a few validation tests to look for resolution of the specific issues DAP brought up. 
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1531)

## This PR
- I have added 3 new tests to test the logic around those recent changes.
- Note that we should be checking the "de_past_residence_country1_text" field for the test_combined_past_countries test but we could not find a valid record to use as a test case.
Check with DAP folks on if that is possible.

### DIFF
--- a/hack/validation/test_release.py
+++ b/hack/validation/test_release.py
@@ -29,7 +29,7 @@ class ReleaseValidationTestCase(unittest.TestCase):
                           msg=f"Row ID {row['entity:cslb_id']} does not have a 2020 cslb_date")
 
     def test_spay_or_neuter_age(self):
-        for row in self._hles_dog2_data:
+        for row in self._hles_dog1_data:
             self.assertIn('dd_spay_or_neuter_age', row)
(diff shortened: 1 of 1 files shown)


raaidbroad (approved):

hack/validation/test_release.py
```diff
@@ -29,7 +29,7 @@ def test_cslb_date(self):
                           msg=f"Row ID {row['entity:cslb_id']} does not have a 2020 cslb_date")
 
     def test_spay_or_neuter_age(self):
-        for row in self._hles_dog2_data:
+        for row in self._hles_dog1_data:
```
aherbst-broad, line 32: 🙈 out of scope but: we need to get rid of this switcheroo-ing, I'll poke at that this week.

aherbst-broad (approved):

PR #150, opened 2021-03-18
Revise Eye Condition Causes

quazi-h:
## Why
DAP requested a column name change hs_condition_cause to hs_eye_condition_cause because this is the only health condition for which a cause is asked with multi-select options. DAP also requested that we pull hs_dx_eye_cause for dogs with blindness to populate this variable and use a default value of 99 when no cause is provided
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1593)

## This PR
- Renamed TDR variable hs_condition_cause to hs_eye_condition_cause.
- Updated if...else logic for blindness to inject a default value of 99 when a cause is not provided
- Updated unit tests to account for these changes


aherbst-broad (approved):
Looks good; is there a verification plan for this (i.e,. an example record we can verify against)?

ThePletch (approved):
lgtm, echoing drew's question about independent validation

PR #152, opened 2021-03-22
Configure JSON printer to include null fields in HLES/CSLB pipelines

ThePletch:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1661)

We were losing a field in the HLES output because none of the dog records in the survey included a value for it. This change causes the JSON printer to still include null values, so no fields will be dropped under any circumstances. This should only affect the size of the JSON - TSV size will be unaffected other than the addition of this errantly omitted column.

I've left a comment explaining why this change is not being applied to the environment pipeline.

This change depends on the merging and release of https://github.com/DataBiosphere/ingest-utils/pull/19 (and the compilation for this PR will fail until that happens)

### DIFF
--- a/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/EnvironmentTransformationPipelineBuilder.scala
+++ b/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/EnvironmentTransformationPipelineBuilder.scala
@@ -25,6 +25,10 @@ object EnvironmentTransformationPipelineBuilder extends PipelineBuilder[Args] {
     val environment =
       rawEnvRecords.transform("Map Environment")(_.map(EnvironmentTransformations.mapEnvironment))
 
+    // unlike our other pipelines, we don't configure our printer to include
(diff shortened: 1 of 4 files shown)


dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/environment/EnvironmentTransformationPipelineBuilder.scala
```diff
@@ -25,6 +25,10 @@ object EnvironmentTransformationPipelineBuilder extends PipelineBuilder[Args] {
     val environment =
       rawEnvRecords.transform("Map Environment")(_.map(EnvironmentTransformations.mapEnvironment))
 
+    // unlike our other pipelines, we don't configure our printer to include
```
aherbst-broad, line 28: Nice descriptive comment.

aherbst-broad (commented):
Looks like a straightforward change, I'll give it a 👍 when checks pass.

ThePletch:
Closed as won't do after feedback from DAP team

PR #153, opened 2021-03-31
Fix DAP ID Issues (primary keys)

quazi-h:
## Why
In our tsv output script, we are generating and appending some artificial primary keys for the DAP tables. This is done so that we can upload the TSVs via the Firecloud api properly since it requires that each TSV have a first column with the label "[entity]:[table_name]_id". The logic for the generated IDs needed a refactor and in addition to that, we have been delivering TSVs without the generated IDs for the benefit of the DAP team.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1630)

## This PR

- Added an argument to execute branching logic to generate aritificial primary keys as needed for Firecloud uploads.
- Refactored code to generate PKs to address some bugs (ex. 'entity:dog_id_id') brought up by DAP.
- Modified the default logic to maintain existing primary keys and not generate any new ones.
- Updated validation script to accommodate these changes.

### DIFF
--- a/hack/convert-output-to-tsv.py
+++ b/hack/convert-output-to-tsv.py
@@ -56,15 +56,13 @@ class PrimaryKeyGenerator:
         else:
             raise ValueError(f"Unrecognized table: {self.table_name}")
 
-    # this will return the firecloud entity names
     def generate_entity_name(self):
         if self.firecloud:
             return f"{PRIMARY_KEY_PREFIX}:{self.table_name}_id"
         return self.pk_name
 
-    # this will return row[entity_name]
     def generate_primary_key(self, row):
-        # normal processing of IDs - PK is stored as PK
+        # normal processing of IDs - the original primary key is returned
         if not self.firecloud:
             return row.pop(self.pk_name)
         # generate ids for Firecloud compatibility
@@ -77,7 +75,7 @@ class PrimaryKeyGenerator:
             try:
                 congenital_flag = int(congenital_flag)
             except TypeError:
-                log.info(f"Error, 'hs_condition_is_congenital' is not populated in {self.table_name}")
+                log.warning(f"Error, 'hs_condition_is_congenital' is not populated in {self.table_name}")
             return '-'.join([str(component) for component in [row.get('dog_id'), row.get('hs_condition'), congenital_flag]])
         # environment: read + copy dog_id and address fields to concatenate for generated uuid
         elif self.table_name == "environment":
--- a/hack/convert-output-to-tsv.py
+++ b/hack/convert-output-to-tsv.py
@@ -21,6 +21,7 @@ parser = argparse.ArgumentParser(description="Convert the records in a specified
 parser.add_argument('input_dir', metavar='I', help='The bucket prefix to read records from')
 parser.add_argument('output_dir', metavar='O', help='The local directory to write the resulting TSVs to')
 parser.add_argument('table', nargs='*', help="One or more tables to process into TSVs. Processes all tables if none specified.")
+parser.add_argument('--firecloud', action='store_true', help="Use logic to generate primary keys for Terra upload via Firecloud")
 parser.add_argument('--debug', action='store_true', help="Write additional logs for debugging")
 args = parser.parse_args()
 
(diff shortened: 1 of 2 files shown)


hack/convert-output-to-tsv.py
```diff
@@ -77,29 +71,41 @@ def remove_prefix(text, prefix):
...
-                        log.info(f"Error, 'hs_condition_is_congenital' is not populated in {table_name}")
-                    row[entity_name] = '-'.join([str(component) for component in [row.get('dog_id'), row.get('hs_condition'), congenital_flag]])
-                elif table_name == "environment":
-                    primary_key_fields = ['dog_id', 'address_1_or_2', 'address_month', 'address_year']
-                    row[entity_name] = '-'.join([str(row.get(field)) for field in primary_key_fields])
+                # generate ids for Firecloud compatibility
+                if args.firecloud:
+                    # hles_health_condition: read + copy dog_id and hs_condition and hs_condition_is_congenital
+                    # concatenate and write out as entity_name, keep original PK
+                    if table_name == "hles_health_condition":
```
aherbst-broad, line 78: 🔧 I think it's time to refactor this a bit, this is a large conditional with many branches that is difficult to reason about. What if we start by pulling lines 78-92 into a separate method that generates a firecloud formatted row?

aherbst-broad (changes requested):
Looking good, I think there's a bit of refactoring to be done to help maintainability.

hack/convert-output-to-tsv.py
ThePletch: agreed, there's an enormous amount of if statements nested here. i have some ideas for turning this into either a service object or just a function, but that's likely too high-involvement for now.

ThePletch (changes requested):
logic seems sound, but i agree with drew that the sheer size of the functions in this script are beginning to make it difficult to review/reason about. let me know if you want to pair on some refactoring.

hack/convert-output-to-tsv.py
```diff
@@ -36,34 +38,64 @@ def remove_prefix(text, prefix):
 
 gcs = GCSFileSystem()
 
+# create a service object to handle all aspects of generating a primary key
+@dataclass
+class PrimaryKeyGenerator:
```
aherbst-broad, line 43: 👍 I think this is a better encapsulation of that complexity.

aherbst-broad (approved):
lgtm

hack/convert-output-to-tsv.py
```diff
@@ -36,34 +38,64 @@ def remove_prefix(text, prefix):
...
+        # most tables should have "dog_id" as a key
+        if self.table_name in {"hles_dog", "hles_cancer_condition", "hles_health_condition", "environment", "cslb"}:
+            self.pk_name = 'dog_id'
+        # owner table is linked to hles_dog via "owner_id"
+        elif self.table_name == 'hles_owner':
+            self.pk_name = 'owner_id'
+        else:
+            raise ValueError(f"Unrecognized table: {self.table_name}")
+
+    # this will return the firecloud entity names
```
ThePletch, line 59: i think this information is conveyed by the method name, so the comment can be nixed

hack/convert-output-to-tsv.py
```diff
@@ -36,34 +38,64 @@ def remove_prefix(text, prefix):
...
+    def generate_entity_name(self):
+        if self.firecloud:
+            return f"{PRIMARY_KEY_PREFIX}:{self.table_name}_id"
+        return self.pk_name
+
+    # this will return row[entity_name]
```
ThePletch, line 65: i don't think this comment is correct

hack/convert-output-to-tsv.py
```diff
@@ -36,34 +38,64 @@ def remove_prefix(text, prefix):
...
+        # hles_health_condition: read + copy dog_id and hs_condition and hs_condition_is_congenital
+        # concatenate and write out as entity_name, keep original PK
+        if self.table_name == "hles_health_condition":
+            # code to generate the value of primary key should be pulled out
+            # grab the congenital flag and convert to int
+            congenital_flag = row.get('hs_condition_is_congenital')
+            try:
+                congenital_flag = int(congenital_flag)
+            except TypeError:
+                log.info(f"Error, 'hs_condition_is_congenital' is not populated in {self.table_name}")
```
ThePletch, line 80: nit: this should be either a `warning` or an `error` instead of an `info`, since it's an error message

ThePletch (approved):
handful of suggestions about comments and such,  but otherwise i think we're looking good here. thanks for your work on this refactor.

PR #156, opened 2021-04-06
Skeleton for sample extraction pipeline

eparkmonster:
## Why
Need to create pipeline for Gencove sample data.
This is a skeleton since we are waiting on filter directives. (noted as TODOs)
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1518)

## This PR

creates the skeleton for the pipeline

### DIFF
--- a/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
+++ b/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
@@ -1,0 +1,50 @@
+package org.broadinstitute.monster.dap.sample
+
+import org.broadinstitute.monster.common.{PipelineBuilder, ScioApp}
+import org.broadinstitute.monster.dap.common._
+
+// Ignore IntelliJ, this is used to make the implicit parser compile.
+import Args._
+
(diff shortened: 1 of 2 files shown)


dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
```diff
@@ -0,0 +1,52 @@
...
+
+object SampleExtractionPipeline extends ScioApp[Args] {
+
+  val forms = List(
+    "dna_kit_tracker",
+    "sample_kit_tracker"
+  )
+
+  // Magic marker for "completed".
+  // NB: We are purposefully excluding the recruitment_fields_complete -> 2
```
aherbst-broad, line 17: Does this comment still apply here?

aherbst-broad: If not, let's remove it as it will lead to confusion in the future.

aherbst-broad (approved):
Looks straightforward

quazi-broad (approved):
Looks good to me, should be pretty quick to update this later when we have the filter logic from the DAP team.

PR #161, opened 2021-04-27
[DSPDC-1711] Create toplevel condition dummy records when needed

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1711)
There are records in DAP where a top level health condition has been filled in, but no specific condition has been entered. In those instances, we should emit a dummy record with the dog ID, the condition type and whether it is congenital or not.

## This PR
* Creates 2 "fill in" lists after computing the set of health conditions that will place dummy records in the output

### DIFF
--- a/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/common/HealthTransformations.scala
+++ b/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/common/HealthTransformations.scala
@@ -7,8 +7,6 @@ import org.broadinstitute.monster.dap.healthcondition.{
 }
 import org.broadinstitute.monster.dogaging.jadeschema.table.HlesHealthCondition
 
-import scala.collection.immutable
-
 object HealthTransformations {
 
   /** Parse all health-condition-related fields out of a raw RedCap record. */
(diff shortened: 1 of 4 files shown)


dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/common/HealthTransformations.scala
```diff
@@ -105,15 +110,52 @@ object HealthTransformations {
...
+    isCongenital: Boolean,
+    keyGetter: (HealthConditionType) => Option[HealthConditionKey]
+  ): Iterable[HlesHealthCondition] = {
+    HealthConditionType.values
+      .filterNot(h => base.exists(c => c.hsConditionType == h.value) || keyGetter(h).isEmpty)
+      .filter(h => rawRecord.getBoolean(keyGetter(h).get.categoryGate))
+      .map { hc =>
+        createHealthConditionRow(
+          rawRecord,
+          keyGetter(hc).get.dataPrefix,
```
aherbst-broad, line 145: I don't like this `get` here, but we should be protected from an error by the filter above: (`keyGetter(h).isEmpty`)

quazi-h (approved):
Looks good to me.
I think I understand the logic within the createFillInConditionRecords function you created and I believe these changes resolve the issues DAP is experiencing

eparkmonster (approved):

PR #163, opened 2021-06-02
Create sample extraction pipeline and test

eparkmonster:
## Why
Need to create extraction pipeline for genetic Sample data.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1516)

## This PR
First half of ticket which is to create the extraction pipeline along with test it. Also edits the sample lookup json schema to reflect the columns that we want to pull.

(Next half, in a different PR will create the transformation pipeline)

Also removed a test in EnvironmentExtractionPipelineSpec.scala that was testing a dynamically generated list of arms against a hard coded one.

### DIFF
--- a/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
+++ b/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
@@ -3,36 +3,41 @@ package org.broadinstitute.monster.dap.sample
 import org.broadinstitute.monster.common.{PipelineBuilder, ScioApp}
 import org.broadinstitute.monster.dap.common._
 
+import java.time.{OffsetDateTime, ZoneOffset}
+
 // Ignore IntelliJ, this is used to make the implicit parser compile.
 import Args._
 
 object SampleExtractionPipeline extends ScioApp[Args] {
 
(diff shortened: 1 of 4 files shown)


dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/sample/SampleExtractionPipeline.scala
```diff
@@ -3,36 +3,41 @@ package org.broadinstitute.monster.dap.sample
...
 
+import java.time.{OffsetDateTime, ZoneOffset}
+
 // Ignore IntelliJ, this is used to make the implicit parser compile.
 import Args._
 
 object SampleExtractionPipeline extends ScioApp[Args] {
 
+  // january 1, 2018 - we ignore any records before this by default (though there shouldn't be any)
+  val Samplepoch = OffsetDateTime.of(2018, 1, 1, 0, 0, 0, 0, ZoneOffset.ofHours(-5))
```
aherbst-broad, line 14: ⛏️ `SampleEpoch` perhaps?

aherbst-broad (approved):
lgtm

quazi-broad (approved):
It's looking pretty good to me, I was going ask about a SampleExtractionPipelineSpec test but I forgot samples is similar to hles where we are only pulling the "baseline_arm_1" event so there shouldn't be any difference to the list of arms from run to run.

PR #164, opened 2021-06-02
Refactor DAP beam runners and import dagster_utils

quazi-h:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1781)
dagster-utils already contains beam runners for dataflow and local, we should make those generic and consumable by DAP, then shift DAP to use those.

## This PR

Parameterizing scala_project and target_class as arguments to LocalBeamRunner (DAP)
- Removed references to scala_project and target_class from the solids code (hardcoded)
- Threaded them through the specific solid configs and pulling them out in the resources as needed
- No longer including them along with the "run_flags" passed to local beam runner and popping them out within resources
- We are now pulling them out of the solid configs passed in at runtime and pulling them out from context in the resources code
- Updated the test_pipelines script to pass the distinct values for those two items to the specific configured solids.


quazi-broad:
PLAN
-Update DAP resources to parameterize scala_project and target_class
-Refactor dagster_utils Beam.py
 - Incorporate DAP beam runner changes to dagster_utils runners (will break HCA)
 - Split into multiple files (Beam.py getting too big)
 - Bump dagster utils version
 - Import and use dagster_utils runners for DAP and :axe: DAP runner resources

quazi-broad:
Initial CI failure (Validate Python) because dap-orchestration depends on broad-dagster-utils (^0.5.0) which doesn’t match any versions, version solving failed. Just merged the dagster_utils PR that this depends on, will rerun checks.

orchestration/pyproject.toml
```diff
@@ -7,8 +7,8 @@ authors = ["Quazi Hoque <qhoque@broadinstitute.org>"]
 [tool.poetry.dependencies]
 python = "~3.9"
 dagster = "^0.11.2"
-broad_dagster_utils = "^0.4.0"
 gcsfs = "^2021.5.0"
+broad-dagster-utils = "^0.5.0"
```
quazi-broad, line 11: I keep getting the CI error:
`Because dap-orchestration depends on broad-dagster-utils (^0.5.0) which doesn't match any versions, version solving failed.` yet I see that [broad-dagster-utils](https://pypi.org/project/broad-dagster-utils/) is on v0.5.0. Am I not using the correct syntax here?

orchestration/dap_orchestration/config/resources/dataflow_beam_runner/global.yaml
```diff
@@ -1,3 +1,5 @@
+# todo: hardcode this to be relative to the solid that’s configuring it by getting the dir from the configured solid
+working_dir: ../../
```
quazi-broad, line 2: @aherbst-broad You mentioned that I should be configuring this path relative to the solid that is configuring it by getting the dir from the configured solid via __file__.

Would I just be doing that in the test_pipelines code or any scratch files I create for testing?
Here's [the example you linked](https://github.com/DataBiosphere/hca-ingest/blob/f6703e4f48dc3a13642d1ffe1a48ba4a14e3da13/orchestration/dagster_orchestration/hca_orchestration/tests/test_pipelines.py#L14-L16
test_pipelines.py)

orchestration/pyproject.toml
aherbst-broad: It might be a race condition with the publishing of the new version; dagster-utils publishes after merge to master so there is always some lag before it's available to pull down. I suggest retrying the failing check.

orchestration/dap_orchestration/config/resources/dataflow_beam_runner/global.yaml
aherbst-broad: We'd need to do it in code, not via YAML. I'd need to think about how to that will play with our `preconfigure` mechanism. I think this is fine for now.

aherbst-broad (approved):
lgtm

orchestration/dap_orchestration/solids.py
```diff
@@ -21,47 +22,52 @@ def extract_records(context: AbstractComputeExecutionContext) -> None:
...
 @configured(extract_records)
 def hles_extract_records(config: dict[str, str]) -> dict[str, str]:
     return _build_extract_config(
-        config,
-        "org.broadinstitute.monster.dap.hles.HLESurveyExtractionPipeline",
-        "raw"
+        config=config,
+        output_prefix="raw",
+        target_class="org.broadinstitute.monster.dap.hles.HLESurveyExtractionPipeline",
+        scala_project="dog-aging-hles-extraction"
```
raaidbroad, line 50: small nit but if the scala project is the same for all these, we can pull it into a `project` variable so that if we ever have to change it we only need to change it in one place

orchestration/dap_orchestration/solids.py
```diff
@@ -21,47 +22,52 @@ def extract_records(context: AbstractComputeExecutionContext) -> None:
...
 
 @configured(extract_records)
 def hles_extract_records(config: dict[str, str]) -> dict[str, str]:
     return _build_extract_config(
-        config,
-        "org.broadinstitute.monster.dap.hles.HLESurveyExtractionPipeline",
-        "raw"
+        config=config,
+        output_prefix="raw",
+        target_class="org.broadinstitute.monster.dap.hles.HLESurveyExtractionPipeline",
```
raaidbroad, line 49: small nit: similar to the scala project, if the org.broadinstitute.monster.dap prefix is the same for target classes over and over again, might as well pull it out into a variable

raaidbroad (approved):
Very nitpicky nits that aren't important, otherwise looks good

orchestration/dap_orchestration/solids.py
quazi-broad: appreciate all the nits!

quazi-broad:
Abstracted some common variables in response to PR comments. Giving my E2E scratch run another go just to be sure.

PR #167, opened 2021-06-09
Add sample to refresh and tsv scripts DAP

eparkmonster:
## Why
Sample extraction and transformation pipeline have been merged. We need to add sample to the refresh and tsv scripts

## This PR
Adds sample to manual refresh and convert TSV script


hack/manual_refresh.sh
```diff
@@ -111,6 +111,11 @@ if [ "$(pipeline_requested environment)" -eq 1 ]; then
 	sbt "dog-aging-hles-extraction/runMain org.broadinstitute.monster.dap.environment.EnvironmentExtractionPipeline --apiToken=$env_automation --pullDataDictionaries=false --outputPrefix=gs://$TARGET_BUCKET/weekly_refresh/$REFRESH_SUBDIRECTORY/raw --runner=dataflow --project=broad-dsp-monster-dap-dev --region=us-central1 --workerMachineType=n1-standard-1 --autoscalingAlgorithm=THROUGHPUT_BASED --numWorkers=4 --maxNumWorkers=8 --experiments=shuffle_mode=service$START_TIME$END_TIME"
 fi
 
+#   SAMPLE
+if [ "$(pipeline_requested sample)" -eq 1 ]; then
```
aherbst-broad, line 115: 🔧 Let's remove this script altogether, we should not be using it anymore.

aherbst-broad (changes requested):
I think we should stop updating the manual refresh script (in fact we should remove it). Instead all refresh should go through the dagster pipeline.

quazi-h (approved):
lgtm

aherbst-broad (approved):

PR #172, opened 2021-06-15
[DSPDC-1771] Add explicit inter-solid dependencies

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1771)
As part of my spiking for the TDR ticket, I discovered we had no explicit dependencies between solids in our refresh pipeline definition. This means that solids can effectively execute in any order and potentially in parallel. We have not bumped into this in practice because dagster runs the solids in the order in which they're declared and serially, when run locally.

## This PR
* Introduces explicit solid dependencies between the extract + transform pairs, and a final "fan-in" to the `write_outfiles` solid. The fan-in consumes a list of results that gets dropped on the floor, but is necessary for the dependency to get created.
* Adds a test case that runs the pipeline using mock resources. I have found this sort of test useful to verify the overall pipeline definition, especially when there are complex fan-ins/outs.
* Shifts our tests to be pure pytest
* Adds `api_token`, `pull_data_dictionaries`, and `end_time` as explicit config schema entries for the extract solids. I am not sure how this ever worked, but I was unable to run the pipeline without these schema items defined.

### DIFF
--- a/orchestration/dap_orchestration/pipelines.py
+++ b/orchestration/dap_orchestration/pipelines.py
@@ -1,4 +1,4 @@
-from dagster import pipeline, ModeDefinition
+from dagster import pipeline, ModeDefinition, ResourceDefinition
 
 from dagster_utils.resources.beam.local_beam_runner import local_beam_runner
 from dagster_utils.resources.beam.dataflow_beam_runner import dataflow_beam_runner
@@ -36,9 +36,18 @@ prod_mode = ModeDefinition(
     }
 )
 
+test_mode = ModeDefinition(
+    name="test",
+    resource_defs={
+        "beam_runner": ResourceDefinition.mock_resource(),
+        "refresh_directory": refresh_directory,
+        "outfiles_writer": ResourceDefinition.mock_resource()
+    }
+)
+
 
 @pipeline(
-    mode_defs=[local_mode, dev_mode, prod_mode]
+    mode_defs=[local_mode, dev_mode, prod_mode, test_mode]
 )
 def refresh_data_all() -> None:
     collected_outputs = [
--- a/orchestration/dap_orchestration/pipelines.py
+++ b/orchestration/dap_orchestration/pipelines.py
@@ -41,14 +41,9 @@ prod_mode = ModeDefinition(
     mode_defs=[local_mode, dev_mode, prod_mode]
 )
 def refresh_data_all() -> None:
-    hles_extract_records()
-    hles_transform_records()
-
-    cslb_extract_records()
-    cslb_transform_records()
-
-    env_extract_records()
-    env_transform_records()
-
-    # the transform output should be in the same directory for all 3 pipelines
-    write_outfiles()
+    collected_outputs = [
+        hles_transform_records(hles_extract_records()),
+        cslb_transform_records(cslb_extract_records()),
+        env_transform_records(env_extract_records())
+    ]
+    write_outfiles(collected_outputs)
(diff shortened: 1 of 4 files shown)


orchestration/dap_orchestration/tests/test_solids.py
```diff
@@ -0,0 +1,157 @@
+import pytest
```
aherbst-broad, line 1: This is the code that was previously in `test_pipelines.py`

orchestration/dap_orchestration/tests/test_pipelines.py
```diff
@@ -1,139 +1,50 @@
-import unittest
```
aherbst-broad, line 1: The solids tests that were here previously have been moved to the `test_solids.py` test case. This file exercises the pipeline e2e.

orchestration/dap_orchestration/solids.py
```diff
@@ -144,10 +166,13 @@ def env_transform_records(config: dict[str, str]) -> dict[str, str]:
         "working_dir": String,
     }
 )
-def write_outfiles(context: AbstractComputeExecutionContext) -> None:
+def write_outfiles(context: AbstractComputeExecutionContext, fan_in_results: list[object]) -> None:
```
aherbst-broad, line 169: Even though it is ignored, the `fan_in_results` input is necessary to express the upstream dependencies.

orchestration/dap_orchestration/pipelines.py
```diff
@@ -36,19 +36,23 @@
...
-    env_transform_records()
-
-    # the transform output should be in the same directory for all 3 pipelines
-    write_outfiles()
+    collected_outputs = [
+        hles_transform_records(hles_extract_records()),
+        cslb_transform_records(cslb_extract_records()),
+        env_transform_records(env_extract_records())
+    ]
+    write_outfiles(collected_outputs)
```
quazi-broad, line 58: this is how we are enforcing the dependency on the upstream steps using the **fan_in_results** param, right?

orchestration/dap_orchestration/solids.py
```diff
@@ -82,7 +103,8 @@ def env_extract_records(config: dict[str, str]) -> dict[str, str]:
         "output_prefix": String,
         "target_class": String,
         "scala_project": String
-    }
+    },
+    input_defs=[InputDefinition("start", Nothing)]
```
quazi-broad, line 107: Is this the piece that allows us to pass the extract functions to the transform ones in the pipelines file?

quazi-h (approved):
This looks good, I think I understand how you enforced an order by adding an extra parameter to the solids:
using input_defs=[InputDefinition("start", Nothing)] for the transform solids 
and fan_in_results: list[object] for the write outfiles, and then just passed then to each other within the pipelines file, is that right?

orchestration/dap_orchestration/solids.py
aherbst-broad: Yes, exactly. This is aka a[ "nothing dependency"](https://docs.dagster.io/concepts/solids-pipelines/pipelines#order-based-dependencies-nothing-dependencies), purely there to indicate ordering to dagster.

orchestration/dap_orchestration/tests/test_solids.py
```diff
@@ -0,0 +1,157 @@
...
+from dagster import ModeDefinition, execute_solid, SolidExecutionResult
+from dagster_utils.resources.beam.noop_beam_runner import noop_beam_runner
+
+import dap_orchestration
+import dap_orchestration.resources
+import dap_orchestration.solids
+
+
+@pytest.fixture
```
raaidbroad, line 10: ah good use of fixtures

raaidbroad (approved):

PR #176, opened 2021-06-23
[DSPDC-1817] Refactor common transformation record grouping code

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1817)
There is much repeated code in our transformation pipeline builders, specifically with respect to reading JSON and transforming to `RawRecord` instances.

## This PR
* Creates a `TransformationHelper` class with necessary grouping semantics
* Adds a test around the methods

## Checklist


quazi-h (approved):
lgtm! This will make it much easier to put up new PipelineBuilders.

I was recently wondering whether we should be grouping the records for all of the pipelines by  redcap_event_name. HLES, Sample, and EOLS are all pulling a single arm (baseline) anyways, so grouping by the redcap event should produce the same output. In particular I was thinking about what changes we will have to make when we have HLES follow up surveys which I think would introduce new arms for us to pull - something Matt has mentioned in the past but I suppose we can handle that pretty easily by just switching the specific function being called when that comes.

dap-etl/transformation/src/test/scala/org/broadinstitute/monster/dap/common/TransformationHelperSpec.scala
```diff
@@ -0,0 +1,39 @@
+package org.broadinstitute.monster.dap.common
+
+import com.spotify.scio.testing.PipelineSpec
+import org.apache.beam.sdk.options.{PipelineOptions, PipelineOptionsFactory}
+import org.scalatest.matchers.should.Matchers
+
+class TransformationHelperSpec extends PipelineSpec with Matchers {
```
quazi-broad, line 7: Actually, I think the spec is failing.

aherbst-broad: Ah, forgot to add the needed fixture data file.

aherbst-broad:
@quazi-broad great point; I think we'll likely want a record per arm globally as you say. Is there any reason _not_ to make that change now?

dap-etl/transformation/src/test/resources/org/broadinstitute/monster/dap/common/transformation_helper/records/records.json
```diff
@@ -0,0 +1,18 @@
+{"record":"11111","redcap_event_name":"jan2020_arm_1","field_name":"study_id","value":"11111"}
```
raaidbroad, line 1: I am having trouble seeing where this is used 👀

dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/common/TransformationHelper.scala
```diff
@@ -0,0 +1,40 @@
...
+          RawRecord(id.toLong, fields + ("redcap_event_name" -> Array(eventName)))
+      }
+  }
+
+  private def readJsonLists(ctx: ScioContext, inputPrefix: String): SCollection[Msg] = {
+    StorageIO
+      .readJsonLists(
+        ctx,
+        "Raw Records",
+        s"${inputPrefix}/records/*.json"
```
raaidbroad, line 36: is this where it is used?

aherbst-broad: @raaidbroad correct. [This ](https://github.com/DataBiosphere/dog-aging-ingest/blob/fd86f096d386d971d6bdfcaee6fbea439de77006/dap-etl/transformation/src/test/scala/org/broadinstitute/monster/dap/common/TransformationHelperSpec.scala#L11) gets the relevant path in the `resources` folder and we by convention always look for a `records/` subfolder underneath.

dap-etl/transformation/src/test/resources/org/broadinstitute/monster/dap/common/transformation_helper/records/records.json
aherbst-broad: I commented below, but here is where it's [used](https://github.com/DataBiosphere/dog-aging-ingest/blob/fd86f096d386d971d6bdfcaee6fbea439de77006/dap-etl/transformation/src/test/scala/org/broadinstitute/monster/dap/common/TransformationHelperSpec.scala#L11)

raaidbroad (approved):

PR #178, opened 2021-06-25
EOLS new Condition transformation DAP pipeline

eparkmonster:
## Why
We need to extend Eols to have newConditions data
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1802)

## This PR
Adds the newConditions transformation code
Added a unit test to test it
Updated schema and eols transformations
## Checklist

### DIFF
--- a/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
+++ b/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
@@ -5,7 +5,7 @@ import org.broadinstitute.monster.dogaging.jadeschema.fragment.EolsNewCondition
 
 object NewConditionTransformations {
 
-  def mapNewConditionsMetadata(rawRecord: RawRecord): EolsNewCondition = {
+  def mapNewConditionMetadata(rawRecord: RawRecord): EolsNewCondition = {
     //    val ToxinConsumption = Some(diagnosisType.contains("2"))
     val diagnosisType = rawRecord.fields.get("eol_dx")
     val InfectiousDisease = diagnosisType.map(_.contains("1"))
--- a/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
+++ b/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
@@ -7,7 +7,7 @@ object NewConditionTransformations {
 
   def mapNewConditionsMetadata(rawRecord: RawRecord): EolsNewCondition = {
     //    val ToxinConsumption = Some(diagnosisType.contains("2"))
-    val diagnosisType = rawRecord.get("eol_dx")
+    val diagnosisType = rawRecord.fields.get("eol_dx")
     val InfectiousDisease = diagnosisType.map(_.contains("1"))
     val ToxinConsumption = diagnosisType.map(_.contains("2"))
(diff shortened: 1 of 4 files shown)


dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
```diff
@@ -0,0 +1,177 @@
+package org.broadinstitute.monster.dap.eols
+
+import org.broadinstitute.monster.dap.common.RawRecord
+import org.broadinstitute.monster.dogaging.jadeschema.fragment.EolsNewCondition
+
+object NewConditionTransformations {
+
+  def mapNewConditionMetadata(rawRecord: RawRecord): EolsNewCondition = {
+    //    val ToxinConsumption = Some(diagnosisType.contains("2"))
```
quazi-broad, line 9: Eric and I were trying to think about whether there is a quick way to make this code a little more DRY but I don't know if it would be worth the effort given the number of fields in this fragment. I feel like there are a few instances throughout DAP where a list of items is mapped to specific integers and those integers/item prefixes are reused in follow up questions. In particular I thought of the Activity Values in [DemographicsTransformations](https://github.com/DataBiosphere/dog-aging-ingest/blob/master/dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/dog/DemographicsTransformations.scala#L243

quazi-h (approved):
lgtm, approving since we ran this code against extracted EOLS data successfully. Left a comment asking about whether we could possibly DRY up the transformation code but I think it would also be fine as is.

dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
aherbst-broad, line 9: ⛏️ Remove this commented out code.

dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
```diff
@@ -0,0 +1,177 @@
...
+    val diagnosisType = rawRecord.fields.get("eol_dx")
+    val InfectiousDisease = diagnosisType.map(_.contains("1"))
+    val ToxinConsumption = diagnosisType.map(_.contains("2"))
+    val Trauma = diagnosisType.map(_.contains("3"))
```
aherbst-broad, line 13: ⛏️ All of these var names should be `camelCase`, i.e., 
```suggestion
    val trauma = diagnosisType.map(_.contains("3"))
```

dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
aherbst-broad: Yeah, we could whip up a small DSL for this potentially, and then codegen these transformation classes. @quazi-broad can you ticket?

aherbst-broad (approved):
lgtm, there are some comestic issues that need addressing.

dap-etl/transformation/src/main/scala/org/broadinstitute/monster/dap/eols/NewConditionTransformations.scala
aherbst-broad: @quazi-broad another thought occurs, this would be a good coding jamboree project.

PR #186, opened 2021-07-22
Add tests for the transformation pipelines

quazi-h:
## Why
We should add tests for each individual transformation pipeline: hles, cslb, environment, sample, and eols
We should use fixture data as both the consumed inputs as well as the expected outputs to make basic assertions about how the transformation pipeline should behave.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1843)

## This PR
HLES Transformation Pipeline Test + Fixture Data
CSLB Transformation Pipeline Test + Fixture Data
Environment Transformation Pipeline Test + Fixture Data
EOLS Transformation Pipeline Test + Fixture Data
Sample Transformation Pipeline Test + Fixture Data

## Checklist


aherbst-broad (approved):
lgtm, I looked over the fixture data and it seems reasonable.

eparkmonster (approved):
LGTM

PR #190, opened 2021-08-06
Pull redcap tokens from vault DAP

eparkmonster:
## Why
We want to pull redcap token from vault at runtime instead of adding api token individually to each solid config
[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1868)

## This PR
Updates pipelines to require api token as resource def
Removed need to add api token manually for each solid config
Separates logic for using base API token vs environment specific API token
Updates tests to reflect that we don't need to add token to each config
Add bash script to run dev refresh
## Checklist


orchestration/dap_orchestration/resources.py
```diff
@@ -1,10 +1,29 @@
 import os
 import subprocess
 
+from dataclasses import dataclass
 from dagster import Field, resource, StringSource
 from dagster.core.execution.context.init import InitResourceContext
 
 
+@dataclass
+class ApiTokenClass:
```
aherbst-broad, line 10: I think ApiTokenConfig would be a better name

orchestration/dap_orchestration/tests/test_solids.py
```diff
@@ -47,7 +46,8 @@ def mode():
         resource_defs={
             "beam_runner": noop_beam_runner,
             "refresh_directory": dap_orchestration.resources.test_refresh_directory,
-            "outfiles_writer": dap_orchestration.resources.test_outfiles_writer
+            "outfiles_writer": dap_orchestration.resources.test_outfiles_writer,
+            "api_token": ResourceDefinition.mock_resource()
```
aherbst-broad, line 50: This can be a real API config from a file (ie a test.yaml analogous to the global.yaml you introduced);  no reason to have the mock resource here

aherbst-broad (approved):

PR #193, opened 2021-08-31
Output TSVs to GS paths

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1891)
We need to be able to output tsvs to a GS path when running in GCP (as opposed to a local FS path when running locally)

## This PR
* Refactors the convert-output-to-tsv.py script, placing it in the orchestration tree and making it callable from our orch. code directly, rather than having to use `subprocess.run`
* Threads through the dap survey type from `extract` -> `transform` -> `write_outfiles` so we can run subsets of the pipelines as needed (i.e. just `eols` and `sample` for example)

## Checklist


orchestration/dap_orchestration/pipelines.py
```diff
@@ -16,7 +16,8 @@
         "beam_runner": local_beam_runner,
         "refresh_directory": refresh_directory,
         "outfiles_writer": outfiles_writer,
-        "api_token": preconfigure_resource_for_mode(api_token, "local")
+        "api_token": preconfigure_resource_for_mode(api_token, "local"),
+        "io_manager": fs_io_manager
```
aherbst-broad, line 20: This allows us to resume a pipeline run from the failed step, when running locally.

orchestration/dap_orchestration/typing.py
```diff
@@ -0,0 +1,4 @@
+
+
+class DapSurveyType(str):
```
aherbst-broad, line 3: Starting to add some types around all the strings we're passing around between solids.

orchestration/dap_orchestration/pipelines.py
quazi-broad: That will be very helpful!

quazi-broad (approved):
Not a hack script anymore! New changes look good, moving the path checking logic and not having to use subprocess.run makes it look very clean. The new survey type will be very helpful especially for the sample refresh pipeline.

PR #196, opened 2021-09-09
Automate weekly sample refresh runs

quazi-h:
## Why
We need to run the DAP sample refresh on a weekly schedule.
We should have a pipeline to only run sample extract and transform and write outfiles using that transform output.
Upstream changes have already ensured that the transform_records solid returns a DAP_survey_type so we do not need to modify or alter write_outfiles to handle running a single pipeline type.

We MAY need to make changes in order to pass in an additional parameter (--firecloud) to the tsv_convert script so that the correct primary key is used in the final TSV.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/dspdc-1867)

## This PR
Added a new pipeline to exclusively process sample data.
Created a weekly_schedule config for the new sample pipeline.
Added the new pipelines to repositories.
Fixing incorrect RedCap token pull into env variables.
Added a new test for the new sample refresh pipeline.

## Checklist


orchestration/dap_orchestration/pipelines.py
```diff
@@ -75,3 +77,44 @@ def refresh_data_all() -> None:
         eols_transform_records(eols_extract_records())
     ]
     write_outfiles(collected_outputs)
+
+
+@pipeline(
```
aherbst-broad, line 82: I don't think we need a new pipeline. Dagster let's us run subsets of a pipeline via the `solid_selection` arg to this decorator. In this case we can give it `["sample_extract_records*"]`, which will run the "sample_extract_records and all solids downstream of it (i.e., sample_extract_records -> sample_transformation_records -> write_outfiles)

orchestration/dap_orchestration/pipelines.py
```diff
@@ -75,3 +77,44 @@ def refresh_data_all() -> None:
...
+    ]
+    write_outfiles(collected_outputs)
+
+
+@weekly_schedule(
+    pipeline_name="refresh_sample_data",
+    start_date=datetime(2021, 9, 9),
+    execution_time=time(15, 00),
+    execution_timezone="US/Eastern",
+    mode="test_mode",
```
aherbst-broad, line 97: I think the mode name is actually "test"

orchestration/project/build.properties
```diff
@@ -0,0 +1 @@
+sbt.version=1.3.7
```
aherbst-broad, line 1: Is this on purpose?

quazi-broad: I did not explicitly change this - should I back this out?

aherbst-broad: Yeah, please do.

ops/helmfile/dagster/apply.sh
```diff
@@ -62,4 +62,4 @@ else
 fi
 
 helmfile --interactive $COMMAND
-fire_slack_deployment_notification ${ENV} ${GIT_SHORTHASH}
+#fire_slack_deployment_notification ${ENV} ${GIT_SHORTHASH}
```
aherbst-broad, line 65: Uncomment this.

orchestration/workspace.yaml
```diff
@@ -1,2 +1,2 @@
 load_from:
-  - python_file: repositories.py
+  - python_file: prod_repositories.py
```
aherbst-broad, line 2: Point this at dev_repositories.py

aherbst-broad (approved):

orchestration/project/build.properties
quazi-broad: leaving this at 1.3.7 as per our discussion

PR #200, opened 2021-09-27
Switch to graphs

aherbst-broad:


quazi-broad (approved):
This should make setting up tests easier

PR #204, opened 2021-10-06
Retry interceptor

aherbst-broad:
## Why

RedCap will error out due to server instability. We should retry in these instances.

## This PR
* Adds a basic retry interceptor that will detect an unsuccessful response and try again, up to 3 times.

## Checklist


quazi-broad (approved):
Should help with Redcap instability

PR #207, opened 2021-10-13
Validate DAP input dates for the extract solids

quazi-h:
## Why
We should validate proper date type ("%Y-%m-%dT%H:%M:%S-%z") that is provided to the extract solids within dagster.
Dates with the improper format will throw an exception downstream in the extraction scala code.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/dspdc-1907)

## This PR
-Added a simple function to check date format that fails with a ValueError.
-Implemented the check within the base_extract_records solid.
-Updated the pipelines test to include the correct date format (failed with the fake string).
-Added a failure test case for the solid to catch the failure.

## Checklist

### DIFF
--- a/orchestration/dap_orchestration/solids.py
+++ b/orchestration/dap_orchestration/solids.py
@@ -137,7 +137,7 @@ def env_extract_records(context: AbstractComputeExecutionContext) -> DapSurveyTy
     arg_dict = {
         "pullDataDictionaries": "true" if context.solid_config["pull_data_dictionaries"] else "false",
         "outputPrefix": f"{context.resources.refresh_directory}/raw",
-        "endTime": context.solid_config["end_time"],
+        "endTime": check_date_format(context.solid_config["end_time"]),
         "apiToken": context.resources.api_token.env_api_token,
 
     }


orchestration/dap_orchestration/tests/test_solids.py
```diff
@@ -40,6 +40,14 @@ def extract_config():
     }
 
 
+@pytest.fixture
+def invalid_extract_config():
```
aherbst-broad, line 44: 👍 nice, I like the use of a fixture here.

orchestration/dap_orchestration/solids.py
```diff
@@ -55,7 +65,7 @@ def base_extract_records(context: AbstractComputeExecutionContext) -> DapSurveyT
     arg_dict = {
         "pullDataDictionaries": "true" if context.solid_config["pull_data_dictionaries"] else "false",
         "outputPrefix": f"{context.resources.refresh_directory}/{context.solid_config['output_prefix']}",
-        "endTime": context.solid_config["end_time"],
+        "endTime": check_date_format(context.solid_config["end_time"]),
```
aherbst-broad, line 68: I think you'll also need to add a check_date_format call to the env_extract_records solid as it doesn't use this base.

PR #212, opened 2021-10-20
Clear dataflow output dirs prior to execution to avoid duplicate data

aherbst-broad:
## Why

If we don't clear dataflow output dirs prior to running again (say, for the same weekly refresh on retry), we'll end up with duplicate data.

## This PR
* Adds code to delete the data in dataflow output directories

## Checklist


quazi-broad (approved):
Changes make sense to me, this will prevent us from having multiple runs' worth of raw outfiles. I'm wondering if we should run the clear_dir function for transform as well since technically a refresh could fail downstream of the transform files being written, too.

PR #213, opened 2021-10-21
Clear out the dataflow subdirs, not the root

aherbst-broad:
## Why

When cleaning out staging areas, we need to clear out the survey specific subdir, not the root output prefix.

## This PR
* Updates the solids to clear out the sub dir, not the root.
## Checklist


aherbst-broad:
@quazi-broad that's correct; there is no real notion of "directories" in GCS, it's all blobs with different naming schemes.

PR #219, opened 2021-11-05
Add upload to firecloud pipeline

aherbst-broad:
## Why

We should have a reusable pipeline for uploading data to Terra, rather than relying on ad-hoc local scripts.

## This PR
* Adds a simple "firecloud" pipeline that uploads TSVs from a source dir to a Terra project

## Checklist


quazi-broad (approved):
Changes look good to me. 2000 lines per chunk should be pretty safe with the dog data, since I was able to get by with the field eng script which chunks by 5000 I believe.

PR #221, opened 2021-11-30
[DSPDC-1955] Fix retrying logic, add test

aherbst-broad:
## Why

[Relevant ticket](https://broadinstitute.atlassian.net/browse/DSPDC-1955)
Our retrying logic is busted; it doesn't actually retry but instead fails out immediately with an `IllegalStateException`:
```
java.lang.IllegalStateException: cannot make a new request because the previous response is still open: please call response.close()"
```

## This PR
* Closes the response prior to proceeding with the retry chain in the failure case
* Adds a test around this code (😢 which should have been there in the first place)
* Fixes the promise retry logic to not throw an Exception but rather fail the promise with an exception instane

## Checklist
- [x] Documentation has been updated as needed.

### DIFF
--- a/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/common/OkWrapper.scala
+++ b/dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/common/OkWrapper.scala
@@ -11,11 +11,14 @@ import java.io.IOException
 import java.time.Duration
 import scala.concurrent.{Future, Promise}
 
+object OkWrapper {
+  val MAX_RETRIES = 3
+}
+
 class OkWrapper extends HttpWrapper {
   /** Timeout to use for all requests to production RedCap. */
   private val timeout = Duration.ofSeconds(300)
   private val logger = LoggerFactory.getLogger(getClass)
-  private val MAX_RETRIES = 3
 
   /** Construct a client instance backed by the production RedCap instance. */
   def client =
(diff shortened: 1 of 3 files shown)


dap-etl/extraction/src/main/scala/org/broadinstitute/monster/dap/common/OkWrapper.scala
```diff
@@ -44,11 +47,12 @@ class OkWrapper extends HttpWrapper {
         val request = chain.request()
         var response = chain.proceed(request)
         var tryCount = 0
-        while (!response.isSuccessful && tryCount < MAX_RETRIES) {
+        while (!response.isSuccessful && tryCount < OkWrapper.MAX_RETRIES) {
           logger.warn(
             s"Unsuccessful response, retrying [url=${request.url()}, code=${response.code()}, try=${tryCount}]"
           )
           tryCount += 1
+          response.close()
```
aherbst-broad, line 55: This is the crux of the issue, we need to close before proceeding on with the chain. I can confirm that the behavior repros in the supplied test when this line is removed.

dap-etl/extraction/src/test/scala/org/broadinstitute/monster/dap/common/OkWrapperSpec.scala
```diff
@@ -0,0 +1,52 @@
...
+import upack.Int64
+
+class OkWrapperSpec extends FixtureAsyncFlatSpec {
+
+  behavior of "OkWrapper"
+
+  type FixtureParam = MockWebServer
+
+  override def withFixture(test: OneArgAsyncTest) = {
+    val server = new MockWebServer
```
aherbst-broad, line 16: This is provided by okhttp and let's us setup  full mock web server, similar to our redcapclient mocking (if we had the time, I'd say we should scrap our homegrown implementation and move to this).

quazi-broad (approved):
Logic looks good to me, I think it would probably be worth redcapclient mock code with the MockWebServer provided by okhttp.

PR #224, opened 2021-12-14
Force TSV Column Order + Process AfusOwner Data

quazi-h:
## Why
We would like to maintain the order of the original data model (schema) for columns in the final TSVs.
We also need to process afus_owner data and generate a TSV.
[Relevant ticket](https://broadinstitute.atlassian.net/browse/1958)

## This PR
Extends the tsv_convert script to process afus_owner.
Forces the column order to hardcoded lists for tsv column order.

## Checklist


orchestration/dap_orchestration/tsv_convert.py
```diff
@@ -36,6 +37,28 @@ def remove_prefix(text: str, prefix: str) -> str:
 
     return text
 
+# DATA MODELS - Correct Ordering of Columns
+def getColumnOrder(table_name):
+    if (table_name == "cslb"):
+        column_order = ['dog_id', 'cslb_date', 'cslb_year', 'cslb_pace', 'cslb_stare', 'cslb_stuck', 'cslb_recognize', 'cslb_walk_walls', 'cslb_avoid', 'cslb_find_food', 'cslb_pace_6mo', 'cslb_stare_6mo', 'cslb_defecate_6mo', 'cslb_food_6mo', 'cslb_recognize_6mo', 'cslb_active_6mo', 'cslb_score', 'cslb_other_changes']
```
aherbst-broad, line 43: 🔧 I think these column lists are going to be unmaintainable. They should be externalized to a file and parsed from there. I also worry a bit about what happens if there are new columns etc., and how we remember to keep these updated. 

We may be able to kill two birds with one stone by using the schema files as the source of truth for column ordering and parsing those instead. The tricky bit there is those are external to the dagster docker container context and we'd need to figure out a way to get those included in the built image.

aherbst-broad (changes requested):
I think we need to think about the column lists.

aherbst-broad (approved):

PR #226, opened 2022-01-19
Added a filter on the samples pipeline for k1_tube_serial

quazi-h:
## Why
When DAP has to redo dna kits for participants, they clear out the old swab id but not the swab date. This results in a handful of records that are missing a value for the k1_tube_serial. We were only checking the swab date and still trying to pull the kit id as a required field which has been causing the sample pipeline to error in production.

[Relevant ticket](https://broadinstitute.atlassian.net/browse/<ticket_id>)

## This PR

- Added a filter.
- Updated tests.

## Checklist


quazi-broad:
Was debating added an exclusive error to the TransformationLog for this and enumerating over more cases but wasn't sure if it was worth it for these errors since this should be a transient issue? I am also refactoring the errors into a common class in another branch so I didn't want to come up against those changes since there are already a few DAP PRs right now.

aherbst-broad (approved):

